PR #1, opened 2018-02-13
fix(starter): remove PreloadAllModules preloading strategy from routing configuration to make lazy-loaded modules to be actually lazy loaded :)

christophercr:
The PreloadingAllModules strategy caused all the modules (including the lazy-loaded) to be preloaded when the application loads (see: https://angular.io/guide/router#how-preloading-works)


SuperITMan:
Seems ok 😊

PR #15, opened 2018-02-23
feat(prettier) Added prettier for git hook and npm run script

tenretC:
Added prettier for node console use
Added prettier-check script to run prettier on all the project code
Added prettier config (.prettierrc and .prettierignore)
Added stylelint-config-prettier and tslint-config-prettier to avoid conflicts between stylelint and prettier, tslint and prettier
Added tslint-check script to display conflicts between prettier and tslint
Removed tslint options conflicting with prettier
Added husky and lint-staged for git hook
Added precommit script to use prettier in git hook thanks to lint-staged and husky


SuperITMan:
A rebase could be nice because there is a lot of commits...

package.json
```diff
@@ -44,5 +44,24 @@
...
     "type": "git",
     "url": "https://github.com/nationalbankbelgium/stark.git"
+  },
+  "devDependencies": {
+    "husky": "0.14.3",
+    "lint-staged": "7.0.0",
+    "prettier": "1.10.2",
+    "stylelint": "8.2.0",
+    "stylelint-config-prettier": "2.1.0",
+    "tslint": "5.8.0",
```
christophercr: I think we should take the chance to upgrade TSLint to the latest version and see if it is still working fine with prettier

package.json
```diff
@@ -44,5 +44,24 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/nationalbankbelgium/stark.git"
+  },
+  "devDependencies": {
+    "husky": "0.14.3",
+    "lint-staged": "7.0.0",
+    "prettier": "1.10.2",
+    "stylelint": "8.2.0",
```
christophercr: We use now version 9.1.1 of Stylelint, could you change this dependency and check if it still works?

We still need to check whether the dependencies to Stylelint and TSLint are to be placed here or in the build package

.stylelintrc
```diff
@@ -1,4 +1,7 @@
 {
+  "extends": [
```
christophercr: I see we just extend the config from prettier. This means that we don't have any conflicts like in tslint?

starter/tslint.json
```diff
@@ -1,56 +1,43 @@
...
+		"use-pipe-transform-interface": true,
+		// General
+		"no-console": [true, "time", "timeEnd", "trace"],
+		"no-string-literal": false,
+		"no-use-before-declare": true,
+		"object-literal-sort-keys": false,
+		"ordered-imports": false,
+		"variable-name": [true, "allow-leading-underscore", "allow-pascal-case", "ban-keywords", "check-format"]
+	}
+	//These rules are handled by Prettier
```
christophercr: If I understand correctly, we will still get the linting warnings/errors in IntelliJ/webpack for these rules right?

package.json
SuperITMan: The goal is to add a hook "precommit" in all the project. For doing that, we need those dependencies in the root.
In any case, we should upgrade the dependencies to the latest versions

PR #16, opened 2018-02-23
feat(prettier) Added prettier for git hook and npm run script

tenretC:
Added prettier for node console use
Added prettier-check script to run prettier on all the project code
Added prettier config (.prettierrc and .prettierignore)
Added stylelint-config-prettier and tslint-config-prettier to avoid conflicts between stylelint and prettier, tslint and prettier
Added tslint-check script to display conflicts between prettier and tslint
Removed tslint options conflicting with prettier
Added husky and lint-staged for git hook
Added precommit script to use prettier in git hook thanks to lint-staged and husky
Modified .editorconfig for json files


.prettierrc
```diff
@@ -0,0 +1,22 @@
...
+"overrides": [
+        {
+            "files": "*.{pcss,scss,css,json}",
+            "options": {
+                "tabWidth": 2,
+                "useTabs":false
+            }
+        },
+        {
+        	"files": "{package.json,.travis.yml}",
```
SuperITMan: Since json files are taken in "files": "*.{pcss,scss,css,**json**}", package.json should not be mentionned here. Could you please remove it ? :)

.prettierignore
```diff
@@ -0,0 +1,5 @@
+.github/
```
dsebastien: .git/ should probably also be ignored

Issue #73, opened 2018-03-01
build: make sure we fully comply with the Angular Package Format (APF)


dsebastien:
cfr Angular package format (APF): https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview 

Things to do we've identified so far:
* [x] add missing properties in package.json for each package
* [x] add missing properties in tsconfig.json for each package
* [ ] sub-packages: add missing sub-packages support in the build.sh script
* [ ] sub-packages: add sub-package for e2e support in stark-core as example and make sure it works. Concrete example: https://github.com/angular/angular/tree/master/packages/core/testing and https://github.com/dherges/ng-packagr#secondary-entry-points
* [ ] add testing folder to each package (and review how it's handled in angular build)

ncedric:
Done

PR #182, opened 2018-03-05
[#73] adapt package.json

ncedric:
Just adapting package.json to follow Angular package format v5.0

### DIFF
--- a/packages/stark-core/package.json
+++ b/packages/stark-core/package.json
@@ -4,7 +4,7 @@
   "module": "./esm5/stark-core.js",
   "es2015": "./es2015/stark-core.js",
   "main": "bundles/stark-core.umd.js",
-  "typings":"stark-core.d.ts",
+  "types": "stark-core.d.ts",
   "description": "Stark - Core",
   "author": "Stark Team",
   "contributors": [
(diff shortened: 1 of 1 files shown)


packages/stark-core/package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "@nationalbankbelgium/stark-core",
   "version": "0.0.0-PLACEHOLDER-VERSION",
+  "module": "./esm5/stark-core.js",
+  "es2015": "./es2015/stark-core.js",
+  "main": "bundles/stark-core.umd.js",
```
SuperITMan: Why is it no "main": "./bundles/stark-core.umd.js" as it is for "module" and "es2015" ?

packages/stark-core/package.json
```diff
@@ -1,6 +1,10 @@
...
+  "typings":"stark-core.d.ts",
```
SuperITMan: Maybe "types": "stark-core.d.ts" could be better than "typings" ?

christophercr: Indeed, "typings" was the previous name of this property, now it should be "types" (although both work fine)

packages/stark-core/package.json
ncedric: @SuperITMan 
I have implemened by following the documentation : 
https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview

Kr,
Cédric

Issue #91, opened 2018-03-02
core: logging


dsebastien:
Integrate logging support and log shipping.
Provide in a dedicated module

* [ ] logging API
* [ ] log shipping (depends on config)

PR #297, opened 2018-04-06
feat(logging): implement STARK_APP_CONFIG token and implement Stark logging in Starter

RobbyDeLaet:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Logging service not yet available in the Starter app.

Issue Number: #91

## What is the new behavior?
Logging service implemented in the Starter app.
Created a "STARK_APP_CONFIG" token to provide the StarkConfig from the client app into Stark.

```
[x] Yes
[ ] No
```

"STARK_APP_CONFIG" token has to be used to inject the AppConfig in components and services.

### DIFF
--- a/starter/src/app/home/home.component.ts
+++ b/starter/src/app/home/home.component.ts
@@ -9,6 +9,8 @@ import {
 	StarkHttpSerializerImpl,
 	StarkHttpService,
 	starkHttpServiceName,
+	StarkLoggingService,
+	starkLoggingServiceName,
 	StarkSingleItemResponseWrapper
 } from "@nationalbankbelgium/stark-core";
 import { Observable } from "rxjs/Observable";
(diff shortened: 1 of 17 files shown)


starter/src/app/home/home.component.ts
```diff
@@ -77,33 +81,31 @@ export class HomeComponent implements OnInit {
...
+				serializer: serializer,
+				retryCount: 4
+			});
 		}
 
 		httpRequest$.subscribe(
 			(response: any) => console.log("---------- SUCCESS", response),
-			(error: StarkHttpErrorWrapper) => console.log("---------- ERROR", error),
-			() => console.log("---------- COMPLETE")
+			(error: StarkHttpErrorWrapper) => this.loggingService.error("---------- ERROR", <any>error),
```
SuperITMan: Why do you cast the error as "any" instead of calling error.toString() or something like this ?

christophercr: Indeed, then we need to change our StarkHttpErrorWrapper so that it extends Error so we can keep the stacktrace and simply pass it to the logging service as is

Issue #119, opened 2018-03-02
core: session


dsebastien:
Integrate Stark session.
Try and isolate in a dedicated module / package.

The behavior should only be enabled for NBB apps

* [x] session configuration
* [x] session initialization (user, user service, user repository, profile loading, ...)
* [x] session timeout handling
* [x] session expired page
* [x] session timeout warning dialog
* [X] logout and logout page

christophercr:
Closed in #876

PR #303, opened 2018-04-12
 feat(session): add actions, entities, reducers and services for session module

tenretC:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
Create session module for Stark-core #119

```
[ ] Yes
[x] No
```

## Other information


packages/stark-core/src/configuration/entities/application/app-config.entity.ts
```diff
@@ -58,28 +59,37 @@ export class StarkApplicationConfigImpl implements StarkApplicationConfig {
 	@autoserialize
 	public loggingFlushResourceName?: string;
 
+	@ValidateIf(StarkValidationMethodsUtil.validateIfDefined)
```
christophercr: This validation should not be used here because this property has a default value is set in the constructor in any case... so this validation will always pass.

packages/stark-core/src/configuration/entities/application/app-config.entity.ts
```diff
@@ -88,6 +98,11 @@ export class StarkApplicationConfigImpl implements StarkApplicationConfig {
 	@autoserialize
 	public keepAliveUrl?: string;
 
+	@ValidateIf(StarkValidationMethodsUtil.validateIfDefined)
```
christophercr: This validation should not be used here because this property has a default value is set in the constructor in any case... so this validation will always pass.

packages/stark-core/src/configuration/entities/application/app-config.entity.intf.ts
```diff
@@ -86,6 +98,12 @@ export interface StarkApplicationConfig {
 	 */
 	keepAliveUrl?: string;
 
+	/**
+	 * Option to disable the keepAlive if it not needed for the application.
```
christophercr: typo "if it not needed" => "if it is not needed"  ;)

packages/stark-core/src/session/reducers/session.reducer.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+	});
+
+	describe("on CHANGE_LANGUAGE_SUCCESS", () => {
+
+		it("should set the session language when state given", () => {
+			// create the initial state object
+			const initialState: StarkSession = session;
+			deepFreeze(initialState); //Enforce immutability
+
+			// // Send the CHANGE_LANGUAGE_SUCCESS action to the sessionReducer
```
christophercr: Double "\\" in comments :p   The same in the other comments in this file

packages/stark-core/src/session/services/session.service.ts
```diff
@@ -0,0 +1,411 @@
...
+
+	public constructor(public store: Store<StarkCoreApplicationState>,
+					   @Inject(starkLoggingServiceName) public logger: StarkLoggingService,
+					   @Inject(starkRoutingServiceName) public routingService: StarkRoutingService,
+					   @Inject(STARK_APP_CONFIG) private appConfig: StarkApplicationConfig,
+					   // FIXME Uncomment when XSRF Service is implemented
+					   // @Inject(starkXSRFServiceName) public xsrfService: StarkXSRFService,
+					   public idle: Idle,
+					   injector: Injector,
+					   // @Inject("$translate") $translate: any) {
```
christophercr: This line can be removed right?

packages/stark-core/src/session/services/session.service.ts
```diff
@@ -0,0 +1,411 @@
...
+			emitXhrResult(xhr);
+		}
+
+		// catch any error raised by the browser while opening the connection. for example:
+		// Chrome "mixed content" error: https://developers.google.com/web/fundamentals/security/prevent-mixed-content/what-is-mixed-content
+		// IE "Access is denied" error: https://stackoverflow.com/questions/22098259/access-denied-in-ie-10-and-11-when-ajax-target-is-localhost
+		try {
+			xhr.open("GET", url, async);
+			// FIXME Uncomment when XSRF Service is implemented
+			// this.xsrfService.configureXHR(xhr);
```
christophercr: You can remove this lne of code, at the end the XSRF service should not be used here because it is a GET request (XSRF only makes sense for state-changing requests (PUT, POST, PATCH and DELETE). So you can remove completely the references to the XSRF service from here.

I already did this change in the old Stark... ;)

packages/stark-core/src/session/services/session.service.ts
```diff
@@ -0,0 +1,411 @@
...
+		pingRequestHeaders = pingRequestHeaders.set(StarkHttpHeaders.NBB_CORRELATION_ID, this.logger.correlationId);
+
+		// FIXME Where does DEVELOPMENT Variable come from ???
+		// if (DEVELOPMENT) {
+			this.fakePreAuthenticationHeaders.forEach((value: string, key: string) => {
+				pingRequestHeaders = pingRequestHeaders.set(key, value);
+			});
+		// }
+		
+		if (typeof this.appConfig.keepAliveUrl !== "string") {
```
christophercr: This if block is not necessary for 2 reasons:

1.- the keepAliveUrl has a class validation (IsURL) so this will never be diferent than a string
2.- in the constructor of this service we validate the config so an undefined keepAliveUrl will be caught there meaning that the application will throw an error before reaching this code

packages/stark-core/src/user/entities/user.entity.ts
```diff
@@ -28,27 +28,27 @@ export class StarkUser implements StarkUserProfile, StarkUserSecurityProfile, St
 	@autoserialize
 	public lastName: string;
 
-	@ValidateIf(StarkValidationMethodsUtil.validateIfDefinedAndNotNull)
+	@ValidateIf(StarkValidationMethodsUtil.validateIfDefined)
```
christophercr: Why we now validate only for "undefined" values?  In fact we need to validate for "undefined" and "null" values because the config file (app-config.json) could have some values set to null...

The same applies for the other validations that were changed...

packages/stark-core/src/configuration/entities/application/app-config.entity.intf.ts
```diff
@@ -60,11 +60,23 @@ export interface StarkApplicationConfig {
...
 	/**
 	 * Enable router logging
 	 */
 	routerLoggingEnabled: boolean;
 
+	/**
+	 * Enable router visualizer. Only in DEV (the router visualizer is not available in PROD)
+	 * Default: false
+	 */
+	routerVisualizerEnabled: boolean;
```
christophercr: This should be optional because it has a default value in the constructor...

Issue #94, opened 2018-03-02
core: NGRX


dsebastien:
Add support for NGRX

* [x] switch to official ngrx packages (finally 🍦)
* [x] switch to official ngrx-store-logger
* [x] switch to official ngrx-store-freeze

dsebastien:
Comment from @christophercr:

To keep in mind:
 Don't declare your reducers as a constant: export const myReducer:ActionReducer<T> = ():T =>
{ ... }
This is not recommended due to the issues caused when using Ahead of Time compilation in Angular2+.
Since we want to be prepared for Angular2+, please declare your reducers as functions so that they are Angular2+ ready.
See these issues for more details:
https://github.com/ngrx/store/issues/190
https://github.com/ngrx/store/pull/306/commits/d56352a18fd3004252d9a41104f02e933f7b2480
https://github.com/ngrx/store/pull/208

PR #310, opened 2018-04-16
fix(ngrx-store): use ngrx store feature selectors in Stark Core modules. Use store-freeze and store-logger meta-reducers in Starter

christophercr:
Solves #94 

## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

```
[X] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

All Stark modules defined the NGX-Store by calling `forRoot`. This causes an exception since `forRoot` should only be used once and only by the Application main module.

Besides that, it is not possible to debug the Application State composition in the Starter app.

Issue Number: #94 

## What is the new behavior?
All Stark modules define the store by calling `forFeature` and provide a [feature store selector](https://github.com/ngrx/platform/blob/master/docs/store/selectors.md#createfeatureselector) as described here: [NGRX-Store API](https://github.com/ngrx/platform/blob/master/docs/store/api.md).

To provide better tooling while defining and implementing the app NGRX store, the [ngrx-store-logger](https://github.com/btroncone/ngrx-store-logger) and the [ngrx-store-freeze](https://github.com/brandonroberts/ngrx-store-freeze) were added to the Starter app.

**IMPORTANT: These store meta-reducers are in fact a developer's choice and not Stark's.**

```
[X] Yes
[ ] No
```


packages/stark-core/src/logging/logging.module.ts
```diff
@@ -1,17 +1,16 @@
...
-import { loggingReducer } from "./reducers/index";
+import { loggingReducer, StarkLoggingState } from "./reducers/index";
 import { StarkLoggingActions } from "./actions/index";
 import { StarkLoggingServiceImpl, starkLoggingServiceName } from "./services/index";
-import { StarkLoggingApplicationState } from "../common/store/starkCoreApplicationState";
 
-const reducers: ActionReducerMap<StarkLoggingApplicationState, StarkLoggingActions> = {
-	starkLogging: loggingReducer
+const reducers: ActionReducerMap<StarkLoggingState, StarkLoggingActions> = {
+	logging: loggingReducer
```
dsebastien: Just curious, why the change of name?
Does this change the store structure? If so then I'd prefer to keep our prefixes.

christophercr: Well, since we define now a "feature" store for every core module, then what we have in the Application Store is something like this:

```javascript
{
  MainAppState :{ }  // root state
  StarkLogging: {  // feature state => with prefix
    logging: ...   // without prefix otherwise it is too verbose having another "starkLogging" inside
  }
  StarkSession: {  // the same principle here
    session: ...
  }
}
```

packages/stark-core/src/logging/services/logging.service.ts
```diff
@@ -77,12 +78,10 @@ export class StarkLoggingServiceImpl implements StarkLoggingService {
 			this.logUrl = this.backend.url + "/" + this.appConfig.loggingFlushResourceName;
 			this.generateNewCorrelationId();
 
-			this.store
-				.select<StarkLogging>((state: StarkLoggingApplicationState) => state.starkLogging)
-				.subscribe((starkLogging: StarkLogging) => {
-					this.starkLogging = starkLogging;
-					this.persistLogMessages();
-				});
+			this.store.pipe(select(selectStarkLogging)).subscribe((starkLogging: StarkLogging) => {
```
SuperITMan: Just a suggestion, could we maybe use starkLoggingSelector instead of selectStarkLogging ? :)
It's weird for me having select(select<>). I would prefer if we had something as select(<>Selector)...

christophercr: Mmm I see, well it depends on which point of view you look at it, in my case I thought of this:

1. normally the selector functions return (select) something from the state and as such, the name of the function is imperative: "selectXXXX"
2. follow the guidelines defined in the NGRX-Store API, since these functions will be exposed, keeping in mind that now any developer can use our framework and we should try not to diverge too much (naming conventions, API, ...) from the rest of the ecosystem. Otherwise my fear is that the learning curve of our framework, which is already important, will deter the devs from adopting it

Issue #96, opened 2018-03-02
core: http


dsebastien:
Add support for NBB's REST design guide with our http api on top of NG5+'s http service
Keep our current model with cerialize.

Should be a dedicated sub-package and module in stark-core

Ref: https://angular.io/guide/http

dsebastien:
blocked by #73

christophercr:
Just to let you know that currently the build of stark-core fails due to these 2 errors:

- the compilation of TS files with decorators from `cerialize` and `class-validator` fails because of this issue https://github.com/angular/angular-cli/issues/8434 so I'm looking at this to see if we can upgrade the angular/cli version.

- the Angular compilar (ngc) throws lots of errors due to the redundant JSDocs we have in some of our classes (see https://github.com/angular/angular/issues/19969). So I have fixed these issues in [this commit](https://github.com/christophercr/stark/commit/1e8a6cf16b77f607eef8dd7af58286cf6a72cfea) in my fork.

PR #318, opened 2018-04-19
feature(error): implement StarkError classes and adapt http and logging service

RobbyDeLaet:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The Error object only provides a message.
When logging errors, a cast is needed for other types of errors, like for instance the HttpError

Issue Number: #96

## What is the new behavior?
The Error object provides: message, name (=error type) , timestamp, correlationId and stack information.
HTTP errors extend the StarkError object.
The logging service accepts Error objects and all error objects that extend the StarkError.
The logging service provides more information.
Developers can create their own error types that extend the StarkError.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/stark-core/src/common/error/error.ts
+++ b/packages/stark-core/src/common/error/error.ts
@@ -1,0 +1,61 @@
+import moment from "moment";
+import { autoserialize } from "cerialize";
+
+import { StarkError } from "./error.intf";
+
+export class StarkErrorImpl implements StarkError {
+	@autoserialize public message?: string;
+	@autoserialize public name?: string;
+	@autoserialize public stack?: string;
+	@autoserialize public correlationId?: string;
+	@autoserialize public timestamp?: string;
+
+	public constructor(error?: Error | string) {
+		let starkError: Error;
+
+		if (!error) {
+			starkError = new Error();
+			this.message = "";
+		} else {
+			if (!(error instanceof Error)) {
+				starkError = new Error(error);
+			} else {
+				starkError = error;
+			}
+
+			this.message = starkError.message;
+		}
+
+		this.name = "STARK_ERROR";
+		this.stack = this.getStackTrace(starkError);
+		this.timestamp = moment().format(); // ISO-8601 format
+	}
+
+	private getStackTrace(error: Error): string {
+		try {
+			if (!error || !error.stack) {
+				error = new Error();
+			}
+
+			let stackTrace: string = <string>error.stack;
+			let index: number;
+
+			// remove the original error message, to keep only the stacktrace
+			if (stackTrace.toLowerCase().startsWith("error")) {
+				index = stackTrace.toLowerCase().indexOf("at");
+				stackTrace = stackTrace.substr(index);
+			}
+
+			// remove own reference from the stacktrace
+			index = stackTrace.toLowerCase().indexOf("starkerrorimpl");
+			if (index >= 0) {
+				index = stackTrace.toLowerCase().indexOf("at", index + 1);
+				stackTrace = stackTrace.substr(index);
+			}
+
+			return stackTrace;
+		} catch {
+			return "";
+		}
+	}
+}
--- a/packages/stark-core/src/common/error/index.ts
+++ b/packages/stark-core/src/common/error/index.ts
@@ -1,0 +1,2 @@
+export * from "./error.intf";
+export * from "./error";
--- a/packages/stark-core/src/http/services/http.service.ts
+++ b/packages/stark-core/src/http/services/http.service.ts
@@ -13,7 +13,7 @@ import { map } from "rxjs/operators/map";
 import { retryWhen } from "rxjs/operators/retryWhen";
 import { mergeMap } from "rxjs/operators/mergeMap";
 import { Injectable } from "@angular/core";
-import { HttpClient, HttpHeaders, HttpResponse } from "@angular/common/http";
+import { HttpClient, HttpHeaders, HttpResponse, HttpErrorResponse } from "@angular/common/http";
 
 import { StarkHttpService, starkHttpServiceName } from "./http.service.intf";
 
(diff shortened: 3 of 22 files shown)


packages/stark-core/src/common/error/error.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      // remove own reference from the stacktrace
+      index = stackTrace.toLowerCase().indexOf("starkerrorimpl");
+      if (index >= 0) {
+        index = stackTrace.toLowerCase().indexOf("at", index + 1);        
+        stackTrace = stackTrace.substr(index);
+      }
+
+      return stackTrace;
+		} catch {
```
christophercr: The formatting is wrong, you can fix it by running prettier manually in all files: `npm run prettier-check`

packages/stark-core/src/common/error/index.ts
```diff
@@ -0,0 +1,4 @@
+"use strict";
```
christophercr: Now we don't need to add this "use strict", it is already taken care by Typescript compiler thanks to the "alwaysStrict" option enabled in the tsconfig.json ;)

Can you remove it from the rest of the files of this PR?

packages/stark-core/src/http/services/http.service.ts
```diff
@@ -313,6 +311,17 @@ export class StarkHttpServiceImpl<P extends StarkResource> implements StarkHttpS
 			httpResponse$ = this.addRetryLogic(httpResponse$, retryCount);
 		}
 
+		let dummyError: Error = new Error(starkHttpServiceName + ": Error getting a SingleItemResponse");
+		if (!dummyError.stack) {
+			// IE 11 won't generate a stack unless the error is thrown
+			// https://docs.microsoft.com/en-us/scripting/javascript/reference/stack-property-error-javascript#remarks
+			try {
+				throw new Error(starkHttpServiceName + ": Error getting a SingleItemResponse");
```
christophercr: I think you can simply throw the same error you defined just above ("dummyError") instead of creating a new error again.

packages/stark-core/src/http/services/http.service.ts
```diff
@@ -313,6 +311,17 @@ export class StarkHttpServiceImpl<P extends StarkResource> implements StarkHttpS
 			httpResponse$ = this.addRetryLogic(httpResponse$, retryCount);
 		}
 
+		let dummyError: Error = new Error(starkHttpServiceName + ": Error getting a SingleItemResponse");
```
christophercr: Could you add a comment to explain why we create this "dummyError" here? This way it will be clear for anyone who sees this code ;)

And perhaps we could rename this error to be more descriptive instead of "dummy..."

packages/stark-core/src/http/services/http.service.ts
```diff
@@ -343,6 +350,17 @@ export class StarkHttpServiceImpl<P extends StarkResource> implements StarkHttpS
 			httpResponse$ = this.addRetryLogic(httpResponse$, retryCount);
 		}
 
+		let dummyError: Error = new Error(starkHttpServiceName + ": Error getting a CollectionResponse");
+		if (!dummyError.stack) {
+			// IE 11 won't generate a stack unless the error is thrown
+			// https://docs.microsoft.com/en-us/scripting/javascript/reference/stack-property-error-javascript#remarks
+			try {
+				throw new Error(starkHttpServiceName + ": Error getting a CollectionResponse");
```
christophercr: I think you can simply throw the same error you defined just above ("dummyError") instead of creating a new error again.

packages/stark-core/src/logging/entities/log-message.entity.intf.ts
```diff
@@ -1,11 +1,12 @@
 "use strict";
 
 import { StarkLogMessageType } from "./log-message-type.entity";
+import { StarkError } from '../../common';
```
christophercr: Are we sure this exports are working fine with Rollup? Could you double check?
As far as I remember we have to add the "/index" in the path so that Rollup can take the index.ts file correctly

packages/stark-core/src/logging/services/logging.service.spec.ts
```diff
@@ -181,7 +182,34 @@ describe("Service: StarkLoggingService", () => {
 			expect((<StarkLogMessage>message).message).toContain("end of message");
 			expect((<StarkLogMessage>message).timestamp).toBeDefined();
 			expect((<StarkLogMessage>message).error).toBeDefined();
-			expect((<StarkLogMessage>message).error).toContain("this is the error");
+
+			const error: StarkError = (<StarkError>(<StarkLogMessage>message).error);
+			if (error) {
```
christophercr: I would suggest to replace this if block by this:

```javascript
expect(error).toBeDefined();  // so the test will fail in case it is undefined
expect((<string>(error).message)).toContain("this is the error");
```

packages/stark-core/src/logging/services/logging.service.spec.ts
```diff
@@ -181,7 +182,34 @@ describe("Service: StarkLoggingService", () => {
...
+			expect(dispatchedAction.type).toBe(StarkLoggingActionTypes.LOG_MESSAGE);
+			const { message }: any = dispatchedAction;
+			expect(message instanceof StarkLogMessageImpl).toBe(true);
+			expect((<StarkLogMessage>message).type).toBe(StarkLogMessageType.ERROR);
+			expect((<StarkLogMessage>message).message).toContain("dummy error message");
+			expect((<StarkLogMessage>message).timestamp).toBeDefined();
+			expect((<StarkLogMessage>message).error).toBeDefined();
+
+			const error: StarkError = (<StarkError>(<StarkLogMessage>message).error);
+			if (error) {
```
christophercr: Same here, we could replace the if block with an expectation (see my previous comment)

packages/stark-core/src/logging/services/logging.service.spec.ts
```diff
@@ -230,7 +257,15 @@ describe("Service: StarkLoggingService", () => {
 				expect(message.message).toContain("dummy error message");
 				expect(message.timestamp).toBeDefined();
 				expect(message.error).toBeDefined();
-				expect(message.error).toContain("this is the error");
+
+				const starkError: StarkError = (<StarkError>message.error);				
+				if (starkError) {
```
christophercr: Same here, we could replace the if block with an expectation (see my previous comment)

Issue #382, opened 2018-05-18
all: implement partial AppConfig validations depending on its usage by Stark services


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Some services in the different Stark modules need the AppConfig entity to be provided but this is not ensured by any means.

## Expected behavior

Some logic in the service constructor should ensure that the injected AppConfig for the different services is present and valid before such services are initiliazed.

## What is the motivation / use case for changing the behavior?

Due to the modularity we want to achieve in Stark, every module should take care of validating the piece of the AppConfig it cares about.

Issue #51, opened 2018-03-01
core: config - revisit


dsebastien:
Revisit stark config files

Do we keep the concept? (vs package.json extensions)
Consider impact on our deploy tool.

dsebastien:
Previous comments (internal JIRA):

Stark's metadata configuration file (stark-app-metadata.json) creates duplication of information with the package.json file (name, description, version).

Moreover the rest of the metadata stored in the configuration file can actually be added to the package.json file instead.

With this jira we should:
- eliminate that file and tell everyone to get rid of it
- document the properties to add to the package.json
- request everyone to add the missing properties to their package.json
- revisit the StarkApplicationMetadata entity
  -> should it now simply be the package.json, allowing to retrieve all the information?
- make it simple to load (same mechanism as before? At the end of the day, just an Angular constant

Some parts of StarkAppConfig could also be moved to package.json (?)

In the package.json we should create a starkApplicationMetadata key to isolate our specific properties

@christophercr : Regarding the "supportedLanguages" option in the metada, I think in fact that this could be part of the application configuration file instead of the metadata / package.json, don't you think?
 -> @dsebastien: Yep indeed

christophercr:
As we've discussed, it would make more sense now to integrate the AppConfig and AppMetadata into the `angular.json` file

PR #383, opened 2018-05-23
feat(core): validate the AppConfig in every service where it is injected

christophercr:
ISSUES CLOSED: #382

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #382 

## What is the new behavior?
All Stark services using the AppConfig, validate such entity (only the props they care about) during initialization.

```
[ ] Yes
[X] No
```

## Other information


christophercr:
I've updated this issue #51 regarding the integration of our current AppConfig and AppMetadata with the `angular.json` file

packages/stark-core/src/modules/settings/services/settings.service.intf.ts
```diff
@@ -1,4 +1,9 @@
+import { InjectionToken } from "@angular/core";
+
 export const starkSettingsServiceName: string = "StarkSettingsService";
```
SuperITMan: Do we still need to export this const ? 
Maybe we should be exhaustive with our export in barrels because I've the feeling this serviceName const is uses only inside the service itself. What do you think about this ?

christophercr: That's true, it is only used inside every service... we could leave them out from the barrels if you want. Could you maybe create an issue for this?

SuperITMan: Thanks 😃

Issue #123, opened 2018-03-02
ui: integrate icons


dsebastien:
Integrate the icons pack and support in stark

PR #423, opened 2018-06-08
feat(ui): integrate @mdi/angular-material icons

christophercr:
ISSUES CLOSED: #123

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #123 

## What is the new behavior?
Stark-UI just ships the mdi icon set and is used by different components in the package.
**If the applciation wants to use this icon set, they should register it to the Angular Material's icon registry, otherwise stark-ui components will have no icons!**

### The rationale behind this is:
- this will give enough freedom to the developers to use any other icon set they may like
- the only restriction in order to not break the stark-ui components is that the icon set they define should contain the icons used by those components (the icon id should be the same). Therefore it will be important to specify which icon (icon id) is used in which component so it is clear what the icon set should contain in case another icon set is used.

```
[ ] Yes
[X] No
```


starter/angular.json
```diff
@@ -19,8 +19,8 @@
             "assets": [
               {
                 "glob": "mdi.svg",
-                "input": "node_modules/@mdi/angular-material",
-                "output": "./assets/icons"
+                "input": "../node_modules/@mdi/angular-material",
```
SuperITMan: Now @mdi/angular-material dependency is part of stark-ui, I think we should move this into angular.json file of stark-ui

showcase/src/app/material-icons.config.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Registers the @mdi svg icon set to the Angular Material icon registry in the default namespace
+ * @param matIconRegistry - Angular Material's icon registry
+ * @param domSanitizer - Angular's DOM sanitizer
+ */
+export function registerMaterialIconSet(matIconRegistry: MatIconRegistry, domSanitizer: DomSanitizer): void {
+	// IMPORTANT: the path of the @mdi svg icon set should match the output path given to it in the angular.json file
+	const iconSetUrl: string = "assets/material-icons/mdi.svg";
+
+	matIconRegistry.addSvgIconSet(domSanitizer.bypassSecurityTrustResourceUrl(iconSetUrl));
```
SuperITMan: What do you think of providing this file directly in Stark-UI?

starter/angular.json
christophercr: Mmm in fact I didn't add it to the angular.json on purpose, why? because if the app doesn't want to add the @mdi icons, they will be included anyways in the /assets folder which will increase the size of their bundle :s and there is no way we can disable the assets from being copied.

Therefore, as I explained in the rationale I mentioned above, they can use whatever icon set they may like and they should add the necessary entry to the angular.json file to copy those to their outputh path, in this case they are also free to decide what the outputh folder will be.

showcase/src/app/material-icons.config.ts
christophercr: The same reason I explained above...

mdi:
Please stop tagging me in this repo's PRs/issues. @mdi is my username.

showcase/src/app/material-icons.config.ts
SuperITMan: But that means our components won't be displayed correctly if they don't have the needed icons... :confused:
We should be careful about that

christophercr: Indeed, I understand but in this case we cannot enforce our icons (mdi icons) since it is a big file and in case they don't want those icons but others we will still copy our icons to the /assets folder.

In fact, the only requirement we have for the stark components to work is that there should be an icon with the same id in the icon set they provide. Therefore I think the most practical solution will be to leave that open to the user and just require them to register the icon set themselves, even if is the mdi icon set shipped in stark-ui.

Issue #456, opened 2018-06-22
ui: theming - theme SCSS import


catlabs:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Expected behavior

Only import the scss theme files of the imported components

Issue #472, opened 2018-06-28
ui: theming - import global styles from old stark


catlabs:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior
In the old stark, we had an extra layer of style above components which take cares of fonts, background etc...

## Expected behavior
We should import those styles and convert the old pcss vars in scss vars. The idea is to reflect the changes of the new Angular Material theme into the old pcss.

PR #473, opened 2018-06-28
feat(stark-ui): import old stark ui global theming

catlabs:
ISSUES CLOSED: #456, #472

## PR Checklist

## PR Type
SCSS restructuration

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```


packages/stark-ui/assets/styles/_base.scss
```diff
@@ -0,0 +1,1653 @@
...
+  flex-direction: column;
+  & > * {
+    flex: 1;
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+md-card {
```
SuperITMan: Is it still relevant ?
Maybe should it be modified by mat-card ?

packages/stark-ui/assets/styles/_base.scss
```diff
@@ -0,0 +1,1653 @@
...
+  font-family: mat-font-family($typography-config);
+  font-size: mat-font-size($typography-config, body-1);
+}
+
+input,
+textarea,
+keygen,
+select,
+button,
+md-card.md-default-theme,
```
SuperITMan: Same about md-\*, we should adapt this for mat-\*

catlabs:
About md-* classes, true those rules won't be applied with the Angular Material 2. As we work now with theming for colors and typography, we could also remove the rules on md-* that are linked to color and typography.

The _base.scss file is rather big, so it will take time to go through all of those md-* rules, what I can do is to keep only the style that I am using for now and we can add rules as we progress in the migration. Therefore we can also have a cleaner _base.scss file.

About changing md-primary to mat-primary, it will then be one more operation when we copy/paste component's pcss to scss. mat-primary isn't a standard way to use Angular Material 2 anyway. I could do it but I do not see the added value.

Let me know what you think about those 2 points :-)

showcase/package.json
```diff
@@ -120,24 +120,27 @@
     "@angular/platform-browser-dynamic": "6.0.4",
     "@angular/platform-server": "6.0.4",
     "@angular/router": "6.0.4",
-    "@nationalbankbelgium/stark-core": "file:../dist/packages-dist/stark-core/nationalbankbelgium-stark-core-10.0.0-alpha.2-2c723cb.tgz",
+    "@nationalbankbelgium/stark-core": "file:../dist/packages-dist/stark-core/nationalbankbelgium-stark-core-10.0.0-alpha.3-1dc6c90.tgz",
     "@nationalbankbelgium/stark-ui": "../packages/stark-ui",
+    "@schematics/angular": "^0.6.8",
```
christophercr: Why do we need schematics for the theming?

showcase/angular.json
```diff
@@ -19,7 +19,7 @@
             "assets": [
               {
                 "glob": "mdi.svg",
-                "input": "../node_modules/@mdi/angular-material",
+                "input": "../node_modules/@nationalbankbelgium/stark-ui/node_modules/@mdi/angular-material",
```
christophercr: I think we need to review this. This change is caused by the change in the tsconfig.json to link the stark-ui in the showcase.

Are we sure this will work in a production build? Otherwise we should revert this change as well as in the tsconfig.json and just keep them in our own fork only in development.

packages/stark-ui/assets/styles/_typography.scss
```diff
@@ -2,30 +2,12 @@
...
-
 @if variable-exists(stark-typography-theme) {
   $typography-theme: map-merge($typography-theme, $stark-typography-theme);
 }
 
 $typography-config: mat-typography-config(
-  $font-family: 'Roboto, "Helvetica Neue", sans-serif',
+  $font-family: map-get($typography-theme, font-family),
   $display-4: mat-typography-level-create(map-get($typography-theme, display-4)),
   $display-3: mat-typography-level-create(map-get($typography-theme, display-3)),
```
christophercr: Could you add on the top of this block a comment with the reference where we can find all these properties?

packages/stark-ui/assets/styles/_variables.scss
```diff
@@ -0,0 +1,237 @@
...
+  border-radius: $border-radius
+);
+
+$table-theme: (
+  border-radius: $border-radius
+);
+
+//font-size, line-height, font-weight
+$typography-theme: (
+  font-family: 'Roboto, "Helvetica Neue", sans-serif',
```
christophercr: The fonts should be replaced by the ones we have in the old stark in `src/ui/css/_base-styles.pcss`:
```css
/********** TYPOGRAPHY **********/
body {
  /* system sans-serif font stack */
  font-family: -apple-system, BlinkMacSystemFont,
  Segoe UI, Roboto, Oxygen,
  Ubuntu, Cantarell, Fira Sans,
  Droid Sans, Helvetica Neue, sans-serif;
  font-size: 13px;
}
```

christophercr:
@SuperITMan Regarding the md-* classes, I agree that we could clean them up and just add them when we need them as we progress.

And regarding the `md-primary` to `mat-primary`, I think that md-primary is still valid, because the notion of primary and accent colors is not really Angular Material related but it is Material Design specifics. And also as @catlabs mentioned, this will be an extra change we need to do when migrating components.

packages/stark-ui/assets/styles/_typography.scss
catlabs: Ok i have created a ticket for that as we will know more about it along the development.

https://github.com/NationalBankBelgium/stark/issues/475

christophercr: :ok:

Issue #494, opened 2018-07-09
showcase: replace HighlightJS by Prism in ExampleViewer component


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

The ExampleViewer component uses the [highlight.js](https://github.com/isagalaev/highlight.js) library to leverage syntax highlighting. However, apart from the fact that there is another library we used to do this in the old Stark ([PrismJS](https://github.com/PrismJS/prism)), the `highlight.js` library causes the following failure on the Travis build in Node 6:

```
ERROR in ./node_modules/highlight.js/lib/languages/sqf.js
Module build failed (from ./node_modules/@angular-devkit/build-optimizer/src/build-optimizer/webpack-loader.js):
RangeError: Maximum call stack size exceeded
    at Object.visitEachChild (/home/travis/build/NationalBankBelgium/stark/showcase/node_modules/@angular-devkit/build-optimizer/node_modules/typescript/lib/typescript.js:56952:28)
    at visitor (/home/travis/build/NationalBankBelgium/stark/showcase/node_modules/@angular-devkit/build-optimizer/src/transforms/wrap-enums.js:53:23)
    at visitNode (/home/travis/build/NationalBankBelgium/stark/showcase/node_modules/@angular-devkit/build-optimizer/node_modules/typescript/lib/typescript.js:56836:23)
    at Object.visitEachChild (/home/travis/build/NationalBankBelgium/stark/showcase/node_modules/@angular-devkit/build-optimizer/node_modules/typescript/lib/typescript.js:57078:46)
    at visitor (/home/travis/build/NationalBankBelgium/stark/showcase/node_modules/@angular-devkit/build-optimizer/src/transforms/wrap-enums.js:53:23)
```

This issue has been reported and it is open for some months already: isagalaev/highlight.js#1685

## Expected behavior

[PrismJS ](https://github.com/PrismJS/prism)is used in the ExampleViewer component to provide syntax highlighting.

## Minimal reproduction of the problem with instructions

Trigger a production build for the showcase app with one of these commands under Node 6:

`npm run build:showcase`
`npm run build:showcase:ghpages`

[...]

PR #503, opened 2018-07-12
feat(stark-ui): implement the Stark-Pretty-Print component

RobbyDeLaet:
ISSUES CLOSED: #494, #496

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The Pretty-Print component is not yet added to the stark-ui package.
The ShowCase  uses highlight-js and angular-highlight-js

Issue Number: #494 #496 

## What is the new behavior?
The Pretty-Print component is added to the stark-ui package.
The Pretty-Print component is used in the example-viewer component
The highlight-js and angular-highlight-js are removed from the ShowCase

```
[ ] Yes
[x] No
```

## Other information


packages/stark-ui/package.json
```diff
@@ -28,7 +28,10 @@
   "dependencies": {
     "@mdi/angular-material": "2.4.85",
     "@types/nouislider": "9.0.4",
-    "nouislider": "11.1.0"
+    "@types/prismjs": "^1.9.0",
```
christophercr: Let's use fixed versions for the moment and avoid the wildcards ;)

packages/stark-ui/package.json
```diff
@@ -28,7 +28,10 @@
...
+    "nouislider": "11.1.0",
+    "pretty-data": "^0.40.0",
```
christophercr: Let's use fixed versions for the moment and avoid the wildcards ;)

packages/stark-ui/package.json
```diff
@@ -28,7 +28,10 @@
...
+    "prismjs": "^1.15.0"
```
christophercr: Let's use fixed versions for the moment and avoid the wildcards ;)

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="stark-pretty-print">
```
christophercr: The class in this element should not be added anymore since we tell Angular to add it via the @HostBinding("class").

Can you remove it and adapt the SCSS styles if needed?

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Component to format and highlight code like HTML, CSS, Typescript...
+ * Can be used to display code examples
+ */
+@Component({
+	selector: "stark-pretty-print",
+	templateUrl: "./pretty-print.component.html",
+	encapsulation: ViewEncapsulation.None
+})
+export class StarkPrettyPrintComponent implements OnChanges, OnInit {
+	public prettyString: string;
```
christophercr: As a convention we should put the `@Input` properties first and then the internal properties...

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -0,0 +1,169 @@
...
+	public highlightingEnabled: boolean;
+
+	/**
+	 * Adds class="stark-pretty-print" attribute on the host component
+	 */
+	@HostBinding("class") public class: string = componentName;
+	/**
+	 * data - The text to be pretty printed
```
christophercr: You don't need to put the name of the property at the beginning, it is already implicit due to the position of this JSDoc.

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -0,0 +1,169 @@
...
+	 */
+	@Input() public data: string;
+
+	/**
+	 * format - The format to be used to pretty print the data string
```
christophercr: You don't need to put the name of the property at the beginning, it is already implicit due to the position of this JSDoc.

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -0,0 +1,169 @@
...
+	 */
+	@Input() public format: StarkPrettyPrintFormat;
+
+	/**
+	 * enableHighlighting (optional) - If true, also highlight the pretty printed string
```
christophercr: You don't need to put the name and the "optional" mode of the property at the beginning, it is already implicit due to the position of this JSDoc and the Typescript code.

packages/stark-ui/src/modules/pretty-print/components/_pretty-print-theme.scss
```diff
@@ -0,0 +1,14 @@
+/* ============================================================================== */
```
christophercr: For the pretty-print component doesn't make sense to have a "theme" scss file. In fact these styles should be in a normal scss file, in this case `pretty-print.component.scss`.

You can ask @catlabs  to explain you the purpose of those 2 different files.

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
```diff
@@ -0,0 +1,706 @@
...
+	 * async beforeEach
+	 */
+	beforeEach(async(() => {
+		return TestBed.configureTestingModule({
+			declarations: [StarkPrettyPrintComponent, TestHostComponent],
+			providers: [
+				{ provide: STARK_LOGGING_SERVICE, useValue: new MockStarkLoggingService() },
+				{ provide: STARK_ROUTING_SERVICE, useClass: MockStarkRoutingService }
+			],
+			schemas: [NO_ERRORS_SCHEMA] // tells the Angular compiler to ignore unrecognized elements and attributes
```
christophercr: Please add to the comment which elements are the ones we don't want to include in the TestingModule

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.html
```diff
@@ -0,0 +1,8 @@
...
+	<div *ngIf="highlightingEnabled">
+		<div [innerHTML]="prettyString"></div>
```
christophercr: Is it secure to bind to the innerHTML property like that?
I mean, what happen if the prettyString is an HTML and it contains for example an <script> tag ?

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
```diff
@@ -0,0 +1,706 @@
...
+			});
+
+			it(shouldHaveInputs, () => {
+				expect(component.data).toBe(rawXmlData);
+				expect(component.format).toBe("xml");
+				expect(component.enableHighlighting).toBeUndefined();
+			});
+
+			it("should nicely format raw XML data", () => {
+				let formattedData: string = component.prettyString;
```
christophercr: Apart from asserting the formatted data from the `prettyString` variable, I think you should also assert the actual content of the HTML `<pre>` element in the same way we did in the old Stark. In this way we test the actual rendering of the component. What do you think?

The same for all the other tests in this file ;)

packages/stark-ui/src/modules/pretty-print/components/_pretty-print-theme.scss
SuperITMan: @RobbyDeLaet Could you please remove this file since it is not useful ? 😊

showcase/src/styles/styles.scss
```diff
@@ -2,3 +2,4 @@
 @import "sidenav_temp";
 @import "~basscss/css/basscss.css";
 @import "~normalize.css/normalize.css";
+@import "~prismjs/themes/prism-okaidia.css";
```
SuperITMan: We should add documentation for this component and mention that we have to add an import to prismjs to make it work. Otherwise, it's not obvious...

RobbyDeLaet: Thru, but I'm waiting for the Showcase update, so that we can start adding pages for each component,

showcase/src/app/shared/example-viewer/example-viewer.component.ts
```diff
@@ -2,13 +2,14 @@ import { Component, OnInit, Input, Inject } from "@angular/core";
 import { HttpErrorResponse } from "@angular/common/http";
 import { FileService } from "./file.service";
 import { STARK_LOGGING_SERVICE, StarkErrorImpl, StarkLoggingService } from "@nationalbankbelgium/stark-core";
-
+/* tslint:disable:no-unused-css */
```
SuperITMan: Why do we need to disable the rule `no-unused-css` ?

SuperITMan:
Could you please remove `import * as hljs from "highlight.js";` from showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts ? 😊 
I think the tests are broken because of that

RobbyDeLaet:
@SuperITMan I know, I lost some changes I did before during a squash.

Issue #506, opened 2018-07-13
bootstrap: Angular production mode cannot be enabled due to AbstractStarkMain implementation


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When running the production build of an app that uses the bootstrap main class from stark-core implemented in #486 , Angular fails to enable the "production mode":
```
Error: Cannot enable prod mode after platform setup.
    at An (application_ref.ts:48)
    at decorateModule (abstract-stark-main.ts:140)
    at e.invoke (zone.js.pre-build-optimizer.js:388)
    at t.run (zone.js.pre-build-optimizer.js:138)
    at zone.js.pre-build-optimizer.js:872
    at e.invokeTask (zone.js.pre-build-optimizer.js:421)
    at t.runTask (zone.js.pre-build-optimizer.js:188)
    at m (zone.js.pre-build-optimizer.js:595)
    at t.invokeTask [as invoke] (zone.js.pre-build-optimizer.js:500)
    at _ (zone.js.pre-build-optimizer.js:1540)
```

## Expected behavior

The Angular production mode should be correctly enabled in the production build of the application.

## Minimal reproduction of the problem with instructions

Use the bootstrap logic provided by stark-core in your `main.browser.ts` like this:

```typescript
class Main extends AbstractStarkMain {
	public constructor(env: StarkEnvironment) {
		super(env);
	}

	public main = (): Promise<any> => {
		console.log("Bootstrapping the App");

		// Bootstrap our Angular app with a top level NgModule
		return (
			platformBrowserDynamic()
				.bootstrapModule(AppModule)
				// the line below adapts the module depending on the environment
				// if you don't like what stark does by default, you can instead do your own customizations through the environment.* files
				// and use environment.customizeAppModule instead
				.then(this.decorateModule)
		);
	};
}

new Main(environment).bootstrap();
```
Create the production build of the app and serve the application by running these commands
- `npm run build:prod`
- `npm run server:prod`

Open the app in the browser and check the console.

PR #508, opened 2018-07-13
fix(stark-core): refactor AbstractStarkMain to enable Angular prod mode before bootstraping the app in production build

christophercr:
ISSUES CLOSED: #506

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #506 

## What is the new behavior?
Angular prod mode correctly enabled in production builds.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
+++ b/packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
@@ -3,7 +3,7 @@ import { disableDebugTools, enableDebugTools } from "@angular/platform-browser";
 import { createNewHosts } from "@angularclass/hmr";
 
 import { StarkMain } from "./stark-main.intf";
-import {StarkEnvironment} from "../environment";
+import { StarkEnvironment } from "../environment";
 
 /**
  * Parent class for bootstrapping Stark applications.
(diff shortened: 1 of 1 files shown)


packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
```diff
@@ -137,7 +143,6 @@ We need great software developers like you! https://jobs.nbb.be
 		// written like this because otherwise "this" will not be captured :)
 		if (this.environment.production) {
 			console.log("Customizing configuration for production!");
-			enableProdMode();
 			disableDebugTools();
```
SuperITMan: Maybe could we move "disableDebugTools" with enableProduction and remove the if condition ? 😊

packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
```diff
@@ -99,6 +99,12 @@ NNNNNNNN         NNNNNNNBBBBBBBBBBBBBBBBB   BBBBBBBBBBBBBBBBB
 We need great software developers like you! https://jobs.nbb.be
 		`);
 
+		if (this.environment.production) {
```
SuperITMan: Could we use ENV === "production" instead of this.environment.production ?
Because we already use ENV === "development" at line 118...

packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
christophercr: Mmm I don't think we can get rid of the if condition, because we need to enable/disable the debug tools depending on the environment. So at least we should have the `if (!this.environment.production)`.

So I think we can just simply leave it like it is to see what we do different in dev and prod

packages/stark-core/src/common/bootstrap/abstract-stark-main.ts
christophercr: Indeed, I'll fix that ;)

PR #510, opened 2018-07-13
refactor(stark-starter): enable UI-Router visualizer only in development. Disable visualizer in showcase

christophercr:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

UI-Router visualizer is always enabled in the Starter and in the Showcase.

## What is the new behavior?
- In the Starter, the UI-Router visualizer is enabled only in DEV.
- In the Showcase, the visualizer is always disabled since it is not needed for the moment. It can be enabled again if needed for development purposes.

```
[ ] Yes
[X] No
```

### DIFF
--- a/showcase/src/app/router.config.ts
+++ b/showcase/src/app/router.config.ts
@@ -1,10 +1,9 @@
 import { UIRouter, Category, StateDeclaration } from "@uirouter/core";
-import { Visualizer } from "@uirouter/visualizer";
 
-function logRegisteredStates(registeredstates: StateDeclaration[]): void {
+function logRegisteredStates(registeredStates: StateDeclaration[]): void {
 	let message: string = "=============  Registered Ui-Router states: ==============\n";
 
-	for (const state of registeredstates) {
+	for (const state of registeredStates) {
 		message += "State : " + state.name;
 		message += " [";
 		message += "parent: " + state.parent;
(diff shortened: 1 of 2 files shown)


showcase/src/app/router.config.ts
```diff
@@ -18,8 +17,10 @@ function logRegisteredStates(registeredstates: StateDeclaration[]): void {
 
 export function routerConfigFn(router: UIRouter): void {
 	router.trace.enable(Category.TRANSITION);
-	router.plugin(Visualizer);
-
+	// Enable UI-Router visualizer here if needed (for development purposes only)
+	// if (ENV === "development") {
+	// 	router.plugin(Visualizer);  // Visualizer should be imported from "@uirouter/visualizer"
```
SuperITMan: Why do you disable it always ?

christophercr: Because it is not needed for us in the Showcase, if we need it to check some transitions it will be only locally in our fork while developing ;)

SuperITMan: That means it should be there in development environment. 
The version we publish is in production mode. That means we could enable it.

christophercr: Nop, because nobody wants to use this in development. If someone wants to use it then he simply needs to uncomment this and that's it, but this should not be commited because it is annoying to have this visualizer all the time in dev

SuperITMan: Okido, thanks for the explanation

Issue #476, opened 2018-06-29
ui: button - implement theming


catlabs:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior
Add a theming for the buttons

## Suggestions
- Having 3 size for buttons: small, medium, large
- Add white color for buttons above a dark background
- Change the buttons in the rewrote components that used them

Issue #222, opened 2018-03-07
doc: document coverage reports and related files


dsebastien:
Explain (in the readmes for now):
* that coverage reports are generated
* where they are generated
* format ...

PR #517, opened 2018-07-17
feat(stark-ui): implement the button's theme

catlabs:
ISSUES CLOSED: #476

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Implementation of the button's theme

Issue Number: #476 

## What is the new behavior?
The old style have been merged and some new ones have been added (white buttons)
The html will be a bit different but it follows the angular material way of declaring a button

### DIFF
--- a/packages/stark-ui/src/modules/action-bar/components/_action-bar-theme.scss
+++ b/packages/stark-ui/src/modules/action-bar/components/_action-bar-theme.scss
@@ -1,37 +1,47 @@
-.stark-action-bar {
-  .mat-icon-button {
-    @include stark-button-color(mat-color(map-get($base-theme, primary-palette)));
-  }
-  &.stark-action-bar-full {
-    background-color: rgba(0, 0, 0, 0.85);
-    color: rgba(255, 255, 255, 0.8);
-    .mat-icon-button {
-      @include stark-button-color(rgba($color: #fff, $alpha: 0.8));
-    }
+@media #{$tablet-query} {
+  .stark-action-bar {
     .action-label {
+      &.primary {
+        color: mat-color(map-get($base-theme, primary-palette));
+      }
+      &.warn {
+        color: mat-color(map-get($base-theme, warn-palette));
+      }
+      &.accent {
+        color: mat-color(map-get($base-theme, accent-palette));
+      }
+      &.success {
+        color: map-get(map-get($button-theme, success), color);
+      }
+      &.alert {
+        color: map-get(map-get($button-theme, alert), color);
+      }
+      &.alt {
+        color: map-get(map-get($button-theme, alt), color);
+      }
+      &.neutral {
+        color: map-get(map-get($button-theme, neutral), color);
+      }
+      &.white {
+        color: #fff;
+      }
       &.disabled {
-        color: rgba($color: #fff, $alpha: 0.25);
+        color: map-get($mat-light-theme-foreground, disabled-text);
       }
     }
   }
 }
 
-@media #{$tablet-query} {
-  .stark-action-bar {
-    color: $md-primary;
-    &.stark-action-bar-full {
-      background: none;
-      color: mat-color(map-get($base-theme, primary-palette));
-      .mat-icon-button:not([disabled]) {
-        @include stark-button-color(mat-color(map-get($base-theme, primary-palette)));
-      }
-      [disabled].mat-icon-button {
-        @include stark-button-color(map-get($mat-light-theme-foreground, disabled-button));
-      }
-      .action-label {
-        &.disabled {
-          color: map-get($mat-light-theme-foreground, disabled-button);
-        }
+@media #{$mobile-only-query} {
+  .stark-action-bar-full {
+    background-color: rgba(0, 0, 0, 0.85);
+    .mat-icon-button {
+      @include stark-button-color(rgba($color: #fff, $alpha: 0.8), $light-primary-text);
+    }
+    .action-label {
+      color: rgba(255, 255, 255, 0.8);
+      &.disabled {
+        color: rgba($color: #fff, $alpha: 0.25);
       }
     }
   }
--- a/packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
+++ b/packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
@@ -4,27 +4,27 @@
 		<div *ngFor="let action of actionBarConfig.actions; trackBy: trackAction" [id]="actionBarId + '-' + action.id"
 		(click)="onClick(action, $event)" class="stark-action-bar--action">
 			<ng-container *ngIf="action.isVisible !== false">
-				<button [ngClass]="action.className" [matTooltip]="action.labelSwitchFunction ? action.labelActivated : action.label" [disabled]="!action.isEnabled" mat-icon-button>
-					<mat-icon starkSvgViewBox [svgIcon]="action.iconSwitchFunction ? action.iconActivated : action.icon"></mat-icon>
+				<button [color]="action.buttonColor ? action.buttonColor:'primary'" [ngClass]="action.className" [matTooltip]="action.labelSwitchFunction ? action.labelActivated : action.label" [disabled]="!action.isEnabled" mat-icon-button>
+					<mat-icon starkSvgViewBox [svgIcon]="action.iconSwitchFunction ? action.iconActivated : action.icon" class="stark-small-icon"></mat-icon>
 				</button>
-				<label class="action-label" [ngClass]="{'disabled': !action.isEnabled}" *ngIf="isExtended" translate>{{ action.label }}</label>
+				<label [class]="'action-label '+ (action.buttonColor?action.buttonColor:'primary')" [ngClass]="{'disabled': !action.isEnabled}" *ngIf="isExtended"> <span translate>{{ action.label }}</span></label>
 			</ng-container>
 		</div>
 	</div>
 
 	<div *ngIf="mode === 'full' || alternativeActions" class="alt-actions">
-		<button class="extend-action-bar" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
-			<mat-icon starkSvgViewBox svgIcon="dots-horizontal"></mat-icon>
+		<button class="extend-action-bar" color="primary" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
+			<mat-icon starkSvgViewBox svgIcon="dots-horizontal" class="stark-small-icon"></mat-icon>
 		</button>
-		<button class="open-alt-actions" mat-icon-button *ngIf="alternativeActions" [matMenuTriggerFor]="menu">
-			<mat-icon starkSvgViewBox svgIcon="dots-vertical"></mat-icon>
+		<button class="open-alt-actions" color="primary" mat-icon-button *ngIf="alternativeActions" [matMenuTriggerFor]="menu">
+			<mat-icon starkSvgViewBox svgIcon="dots-vertical" class="stark-small-icon"></mat-icon>
 		</button>
 		<mat-menu #menu="matMenu" xPosition="before">
 				<div *ngFor="let action of alternativeActions; trackBy: trackAction" mat-menu-item [id]="actionBarId + '-alt-' + action.id" [disabled]="!action.isEnabled"
 				(click)="onClick(action, $event)">
 					<ng-container *ngIf="action.isVisible !== false">
 						<div [ngClass]="action.className">
-							<mat-icon starkSvgViewBox [svgIcon]="action.icon"></mat-icon>
+							<mat-icon starkSvgViewBox [svgIcon]="action.icon" class="stark-small-icon"></mat-icon>
 							<span translate>{{ action.label }}</span>
 						</div>
 					</ng-container>
--- a/packages/stark-ui/src/modules/action-bar/components/action.intf.ts
+++ b/packages/stark-ui/src/modules/action-bar/components/action.intf.ts
@@ -3,42 +3,47 @@
  */
 export interface StarkActionBase {
 	/**
-	 * Path to SVG icon from iconSets to display inside the action button. Ex: "pencil" (optional)
+	 * Color of the button
+	 */
+	buttonColor?: "primary" | "accent" | "warn" | "success" | "alert" | "alt" | "neutral" | "white" | string;
+
+	/**
+	 * Path to SVG icon from iconSets to display inside the action button. Ex: "pencil"
 	 */
 	icon?: string;
 
 	/**
-	 * Text to be displayed as label of the action button (optional)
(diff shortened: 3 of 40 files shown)


showcase/src/app/app.component.html
```diff
@@ -31,7 +42,7 @@
 			</a>
 		</mat-nav-list>
 	</mat-sidenav>
-  <mat-sidenav-content class="px4 py2">
+	<mat-sidenav-content class="px4 py2">
```
christophercr: What are these classes for? Where are they defined?

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
```diff
@@ -4,27 +4,27 @@
 		<div *ngFor="let action of actionBarConfig.actions; trackBy: trackAction" [id]="actionBarId + '-' + action.id"
 		(click)="onClick(action, $event)" class="stark-action-bar--action">
 			<ng-container *ngIf="action.isVisible !== false">
-				<button [ngClass]="action.className" [matTooltip]="action.labelSwitchFunction ? action.labelActivated : action.label" [disabled]="!action.isEnabled" mat-icon-button>
-					<mat-icon starkSvgViewBox [svgIcon]="action.iconSwitchFunction ? action.iconActivated : action.icon"></mat-icon>
+				<button color="primary" [ngClass]="action.className" [matTooltip]="action.labelSwitchFunction ? action.labelActivated : action.label" [disabled]="!action.isEnabled" mat-icon-button>
```
SuperITMan: Why force setting primary color?
Maybe should we provide a new property in the interface to set a color? Which is an optional string and, if it is not set, we can set primary by default. What do you think about this? 😊

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
```diff
@@ -4,27 +4,27 @@
...
 				</button>
 				<label class="action-label" [ngClass]="{'disabled': !action.isEnabled}" *ngIf="isExtended" translate>{{ action.label }}</label>
 			</ng-container>
 		</div>
 	</div>
 
 	<div *ngIf="mode === 'full' || alternativeActions" class="alt-actions">
-		<button class="extend-action-bar" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
-			<mat-icon starkSvgViewBox svgIcon="dots-horizontal"></mat-icon>
+		<button class="extend-action-bar" color="primary" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
```
SuperITMan: Why force setting primary color?
Maybe should we provide a new property in the interface to set a color? Which is an optional string and, if it is not set, we can set primary by default. What do you think about this? :blush:

showcase/src/app/app.module.ts
```diff
@@ -154,7 +154,11 @@ export const metaReducers: MetaReducer<State>[] = ENV !== "production" ? [logger
 		StarkRoutingModule.forRoot(),
 		SharedModule,
 		DemoModule,
-		StarkAppLogoModule
+		StarkAppLogoModule,
```
christophercr: All these modules (except the StarkAppLogoModule) should not be imported here, they are now imported in the Demo module ;)

showcase/src/app/demo/button/button.component.ts
```diff
@@ -0,0 +1,16 @@
...
+	selector: "showcase-demo-button",
+	templateUrl: "./button.component.html",
+	styleUrls: ["./button.component.scss"],
+	encapsulation: ViewEncapsulation.None
+})
+export class ButtonComponent implements OnInit {
+	public constructor(@Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService) {}
+
+	public ngOnInit(): void {
+		console.log("On Init");
```
christophercr: Replace the console.log by the StarkLoggingService since you already injected it ;)

showcase/src/app/demo/index.ts
```diff
@@ -2,3 +2,6 @@ export * from "./action-bar";
 export * from "./example-viewer";
 export * from "./table";
 export * from "./demo.module";
+export * from "./action-bar/action-bar.component";
```
christophercr: This re-export is not needed, This is already done in this barrel: `demo/action-bar/index.ts`

showcase/src/app/demo/index.ts
```diff
@@ -2,3 +2,6 @@ export * from "./action-bar";
...
+export * from "./button/button.component";
+export * from "./example-viewer/example-viewer.component";
```
christophercr: This re-export is not needed. This is already done in this barrel: `demo/example-viewer/index.ts`

showcase/src/app/demo/index.ts
```diff
@@ -2,3 +2,6 @@ export * from "./action-bar";
 export * from "./example-viewer";
 export * from "./table";
 export * from "./demo.module";
+export * from "./action-bar/action-bar.component";
+export * from "./button/button.component";
```
christophercr: This barrel should just re-export everything from `./action-bar`. There, inside the "action-bar" folder there should be a barrel `index.ts` which re-exports all the `.ts` files except the `spec.ts`.

See how it is done in the other folders ;)

packages/stark-ui/src/modules/action-bar/components/_action-bar-theme.scss
```diff
@@ -1,38 +1,23 @@
...
+  .action-label {
+    color: mat-color(map-get($base-theme, primary-palette));
+    &.disabled {
+      color: map-get($mat-light-theme-foreground, disabled-text);
+    }
   }
-  &.stark-action-bar-full {
+}
+
+@media #{$mobile-only-query} {
```
christophercr: In the previous code the media query was "tablet-query" and now is "mobile-only-query"... Is this an intended change? Why? What about tablet?

showcase/src/app/demo/button/button.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, OnInit, Inject, ViewEncapsulation } from "@angular/core";
+import { STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
+
+@Component({
+	selector: "showcase-demo-button",
```
christophercr: I'm not sure these lines are correctly formatted (indentation) ... could you please run the "prettier-check" command to auto format all files?

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
catlabs: This is a feature coming from Angular Material, nothing added on our side.

https://material.angular.io/components/button/examples

showcase/src/app/app.component.html
catlabs: As discussed 2 weeks ago, those classes are coming from basscss. It provides some global classes (padding in this case) which is convenient to avoid having extra css on our side.
Regarding sidenav, we will implement it soon, so this code is going to change.

showcase/src/app/demo/button/button.component.ts
catlabs: oups my mistake :-)

packages/stark-ui/src/modules/action-bar/components/_action-bar-theme.scss
catlabs: Yes it is intended, now with the button theme, we can specify which color the button should have with the color attribute. It is only an small device that we need to use custom colors, so it make more sense to have SCSS exception on small device

showcase/src/app/demo/button/button.component.ts
catlabs: Yes I already did it, so I guess the indentation is correct. Seems ok in my editor...

showcase/src/app/app.component.html
christophercr: Ok

catlabs:
Allright, I repushed the changes. @christophercr @SuperITMan can you recheck?

showcase/src/app/demo/button/button.component.ts
christophercr: Ok

packages/stark-ui/src/modules/action-bar/components/_action-bar-theme.scss
christophercr: Oh I see, Ok.

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
catlabs: If we want a closest integration with Angular Material, I think we should keep the directive that they provide. Now I have extend the 3 default colors to add the one we were using before. Then the end user could use Angular Material button with our custom colors, I think that provide the most flexibility.

Imagine if we give up on the color attr and they write this:
<button color="primary" class="stark-alert" mat-raised-button>Primary or Alert?</button>
Then this button have 2 styles defined, so one should override the other

With this notation:
<button color="alert" mat-raised-button>Alert</button>
There is no confusion and we stay inline with Angular Material

catlabs: If we really want to override the color we can do it through a mixin that I have created, which use actually the angular material mixin for buttons which takes the color and the text color as argument

The code for end user will be:
.my-custom-button {
  @include stark-button-color(#fff, #222);
}

SuperITMan: @catlabs What do you think if we add a property in StarkAction ?
Now, we have something like this:
```typescript
export interface StarkAction {
  // ...
  isEnabled: boolean;
  label: string;
  // ...
}
```

But we could have something else like:
```typescript
export interface StarkAction {
  // ...
  isEnabled: boolean;
  label: string;
  buttonColor: string;    //    <-------
  // ...
}
```

Then, in the component, we could have this:
```html
<button color="action.buttonColor ? action.buttonColor : 'alert'" mat-raised-button>Alert</button>
```

What do you think about this ?
 
We continue to use Angular Material as it should be but we give more abilities to choose the color of the button.
I don't think we should use `@mixin` function to change the color of the button...

catlabs: Well yeah that's an idea, we talked about it with @christophercr actually, I ll do that and repush the change

christophercr: Mmm, but then we should think about what are the possible colors we would accept; any? just primary, accent and alert?

If we accept any, in case a new color is given, then it means that the developer should also "create" that color with the Angular Material mixins which at the end willl be the same effort that the one @catlabs mentioned above in order to override the color.

So, I don't know if this approach of having a buttonColor in the interface will make things easier or not

SuperITMan: I think you're right Christopher :blush:

We could do so
```typescript
{
  //... 
  buttonColor: "primary" | "alert" | "accent";
} 
```

What do you think about this?
And if they want more flexibility, they will have to play with `@mixin` function :blush:

catlabs: Well yeah that's the idea, they can either enter one of the seven colors we provide (primary, warn,accent,success,alert,alt,neutral,white) or a custom one, then they will have to play with the mixin

catlabs:
Changes pushed. @christophercr @SuperITMan can you recheck?

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
```diff
@@ -4,27 +4,27 @@
...
-				<label class="action-label" [ngClass]="{'disabled': !action.isEnabled}" *ngIf="isExtended" translate>{{ action.label }}</label>
+				<label [class]="'action-label '+ (action.buttonColor?action.buttonColor:'primary')" [ngClass]="{'disabled': !action.isEnabled}" *ngIf="isExtended"> <span translate>{{ action.label }}</span></label>
 			</ng-container>
 		</div>
 	</div>
 
 	<div *ngIf="mode === 'full' || alternativeActions" class="alt-actions">
-		<button class="extend-action-bar" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
-			<mat-icon starkSvgViewBox svgIcon="dots-horizontal"></mat-icon>
+		<button class="extend-action-bar" color="primary" mat-icon-button *ngIf="mode === 'full'" (click)="toggleExtendedActionBar()">
```
christophercr: I think this "color" should also be set via the buttonColor right?

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
```diff
@@ -4,27 +4,27 @@
...
+			<mat-icon starkSvgViewBox svgIcon="dots-horizontal" class="stark-small-icon"></mat-icon>
 		</button>
-		<button class="open-alt-actions" mat-icon-button *ngIf="alternativeActions" [matMenuTriggerFor]="menu">
-			<mat-icon starkSvgViewBox svgIcon="dots-vertical"></mat-icon>
+		<button class="open-alt-actions" color="primary" mat-icon-button *ngIf="alternativeActions" [matMenuTriggerFor]="menu">
```
christophercr: Same here, I think this "color" should also be set via the buttonColor right?

packages/stark-ui/src/modules/action-bar/components/action.intf.ts
```diff
@@ -2,6 +2,11 @@
  * Stark Action Base interface
  */
 export interface StarkActionBase {
+	/**
+	 * Color of the button (optional)
+	 */
+	buttonColor?: string;
```
christophercr: The "(optional)" string in the comment is not needed since it is already implicit in the Typescript code. Then, I think the type of this property should be more specific as @SuperITMan suggested, so it would be something like this:

```typescript
/**
 * Color of the action button. Default: "primary".
 */
buttonColor?: "primary" | "alert" | "accent" | "anyOtherStarkColor" | string;
```

In this case, the "string" type is still needed so the developer can give any other color.

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
catlabs: Well this button are not set via any Input property, it is the generic extend button. I don't think we should add an Input for this, it will make the api more complex.

packages/stark-ui/src/modules/action-bar/components/action-bar.component.html
catlabs: Same remark than above

catlabs:
Changes pushed. @christophercr @SuperITMan can you recheck?

SuperITMan:
catlabs could you please rebase your PR? 😊

catlabs:
Rebase pushed

PR #565, opened 2018-07-30
feat(stark-ui): Date Picker Component

catlabs:
ISSUES CLOSED: #542

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #542 

## What is the new behavior?
- Implementation of the Date Picker Component
- Moment translation
- Fix of the showcase header on desktop

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
+++ b/packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
@@ -1,0 +1,15 @@
+<mat-form-field [id]="pickerId">
+	<input matInput 
+		[disabled]="isDisabled"
+		[id] = "pickerId+'-input'"
+		[matDatepicker]="picker" 
+		[placeholder]="label | translate" 
+		[matDatepickerFilter]="dateFilterFnWrapper"
+		[min]="minDate"
+		[max]="maxDate"
+		[name]="pickerName"
+		[value]="date"
+		(dateChange)="onDateChange($event)">
+	<mat-datepicker-toggle matSuffix [for]="picker"></mat-datepicker-toggle>
+	<mat-datepicker [id] = "pickerId" #picker></mat-datepicker>
+</mat-form-field>
--- a/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
+++ b/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
@@ -1,0 +1,139 @@
+import { Component, EventEmitter, ViewEncapsulation, Input, Output, ViewChild } from "@angular/core";
+import { MatDatepicker, MatDatepickerInput, MatDatepickerInputEvent } from "@angular/material";
+import * as moment from "moment";
+
+export type StarkDatePickerFilter = "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined;
+
+/**
+ * Component to display the stark date-picker
+ */
+@Component({
+	selector: "stark-date-picker",
+	templateUrl: "./date-picker.component.html",
+	encapsulation: ViewEncapsulation.None
+})
+export class StarkDatePickerComponent {
+	/**
+	 * Source Date to be bound to the datepicker model
+	 */
+	@Input()
+	public date: Date;
+
+	/**
+	 * Filter function or a string
+	 * Whenever this value is changed, we set the dateFilter
+	 */
+	@Input()
+	public get dateFilter(): StarkDatePickerFilter {
+		return this._dateFilter;
+	}
+	public set dateFilter(value: StarkDatePickerFilter) {
+		this._dateFilter = value;
+		if (this._dateFilter === "OnlyWeekends") {
+			this._dateFilter = this.filterOnlyWeekends;
+		} else if (this._dateFilter === "OnlyWeekdays") {
+			this._dateFilter = this.filterOnlyWeekdays;
+		}
+	}
+	private _dateFilter?: StarkDatePickerFilter;
+
+	/**
+	 * Whether the datepicker is disabled
+	 */
+	@Input()
+	public isDisabled?: boolean;
+
+	/**
+	 * Label to be displayed in the datepicker
+	 */
+	@Input()
+	public label: string;
+
+	/**
+	 * Maximum date of the date picker
+	 */
+	@Input()
+	public maxDate: Date;
+
+	/**
+	 * Minimum date of the date picker
+	 */
+	@Input()
+	public minDate: Date;
+
+	/**
+	 * id attribute of the form field wrapping the mat-datepicker
+	 * id attribute followed by "-input" of the mat-datepicker-input
+	 *
+	 */
+	@Input()
+	public pickerId: string = "";
+	
+	/**
+	 * HTML "name" attribute of the element.
+	 */
+	@Input()
+	public pickerName: string = "";
+
+	/**
+	 * Output that will emit a specific date whenever the selection has changed
+	 */
+	@Output()
+	public dateChanged: EventEmitter<Date> = new EventEmitter<Date>();
+
+	/**
+	 * Reference to the MatDatepicker embedded in this component
+	 */
+	@ViewChild(MatDatepicker)
+	public picker: MatDatepicker<any>;
+
+	/**
+	 * Reference to the MatDatepickerInput embedded in this component
+	 */
+	@ViewChild(MatDatepickerInput)
+	public pickerInput: MatDatepickerInput<any>;
+
+	/**
+	 * Wrap the dateFilter function
+	 * We use the MatMomentDateModule, so the MatDatepicker will return a moment.Moment object. To keep constitency with the old code, the end user should be able to specify a custom dateFilter accepting a Date object as a parameter
+	 * @param momentDate - The date to be checked
+	 * @returns Whether the date is filtered or not
+	 */
+	public dateFilterFnWrapper = (momentDate: moment.Moment): boolean => {
+		const date: Date = momentDate.toDate();
+		if (typeof this._dateFilter === "function") {
+			return this._dateFilter(date);
+		}
+		return true;
+	};
+
+	/**
+	 * Filter only the weekend days
+	 * @param date - The date to be checked
+	 * @returns Whether the date is a weekend day or not
+	 */
+	public filterOnlyWeekends(date: Date): boolean {
+		const day: number = date.getDay();
+		return day === 0 || day === 6;
+	}
+
+	/**
+	 * Filter only the week days
+	 * @param date - The date to be checked
+	 * @returns Whether the date is a week day or not
+	 */
+	public filterOnlyWeekdays(date: Date): boolean {
+		const day: number = date.getDay();
+		return day !== 0 && day !== 6;
+	}
+
+	/**
+	 * Handle the date changed on mat-datepicker and emit the dateChanged event
+	 * @param MatDatepickerInputEvent<Date> - the mat-datepicker event
+	 */
+	public onDateChange(event: MatDatepickerInputEvent<moment.Moment>): void {
+		if (event.value) {
+			this.dateChanged.emit(event.value.toDate());
+		}
+	}
+}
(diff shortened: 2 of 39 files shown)


showcase/src/app/demo/demo.module.ts
```diff
@@ -53,17 +58,21 @@ import {
...
 		ActionBarComponent,
 		ButtonComponent,
+		DatePickerComponent,
 		ExampleViewerComponent,
 		KeyboardDirectivesComponent,
 		PrettyPrintComponent,
 		TableComponent
-	]
+	],
+	providers: [{ provide: MAT_DATE_FORMATS, useValue: STARK_DATE_FORMATS }]
```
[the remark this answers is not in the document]
catlabs: Well it is for sure not trivial cause not intended to be used this way.
The regular way of changing the language is to set the dateAdapter: dateAdapter.setLocale("nl-be");
I checked on the moment documentation and when we format with LL, it displays the date in the set country fashion (here with LLLL):
EN: Friday, June 24, 2016 1:42 AM
DE: Freitag, 24. Juni 2016 01:42

So I think it will be fine :-)

packages/stark-ui/src/modules/date-picker/components/date-picker.config.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Object containing the format of the dates displayed in the date picker
+ */
+export const STARK_DATE_FORMATS: Object = {
```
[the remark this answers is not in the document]
catlabs: same remark as below

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -0,0 +1,131 @@
...
+	 */
+	@Input()
+	public date: Date;
+
+	/**
+	 * Filter function or a string
+	 * Whenever this value is changed, we set the dateFilterFunc
+	 */
+	@Input()
+	public get dateFilter(): "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined {
```
christophercr: You can replace these types by the `StarkDatePickerFilter ` type you just created ;)

packages/stark-ui/src/modules/date-picker/components/date-picker.config.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Object containing the format of the dates displayed in the date picker
```
christophercr: I think we can rename this file to "date-format.constants.ts" since it is just about formats and not a config file.

And can you add the URL where we can find the accepted formats in moment that we can use?

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -0,0 +1,131 @@
...
+	}
+	public set dateFilter(value: "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined) {
+		this._dateFilter = value;
+		if (this._dateFilter === "OnlyWeekends") {
+			this._dateFilter = this.filterOnlyWeekends;
+		} else if (this._dateFilter === "OnlyWeekdays") {
+			this._dateFilter = this.filterOnlyWeekdays;
+		}
+	}
+	private _dateFilter?: "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined;
```
christophercr: You can replace these types by the `StarkDatePickerFilter ` type you just created ;)

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -0,0 +1,131 @@
...
+	public picker: MatDatepicker<any>;
+
+	/**
+	 * Reference to the MatDatepickerInput embedded in this component
+	 */
+	@ViewChild(MatDatepickerInput)
+	public pickerInput: MatDatepickerInput<any>;
+
+	/**
+	 * Wrap the dateFilterFunc function
```
christophercr: Leftover from the previous name: "Func" ;)

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Filter function or a string
+	 * Whenever this value is changed, we set the dateFilterFunc
+	 */
+	@Input()
+	public get dateFilter(): "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined {
+		return this._dateFilter;
+	}
+	public set dateFilter(value: "OnlyWeekends" | "OnlyWeekdays" | ((date: Date) => boolean) | undefined) {
```
christophercr: You can replace these types by the `StarkDatePickerFilter ` type you just created ;)

SuperITMan:
After testing it on my computer, I had some troubles with the date picker...
In the first one, at least, presented in the showcase page, I tested to enter the following dates and I didn't get the attended result:

  - 12/20/2017 returns 12 januari 2020 (attended: 20 December 2017 by guessing EN system)
  - 20/12/2017 returns 20 januari 2012 (attended: 20 December 2017 by guessing FR system)

Could you please check if the locales are correctly set ?

Also, could you please rebase your PR for getting changes in the example-viewer (`title` became `exampleTitle`)

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
```diff
@@ -0,0 +1,14 @@
+<mat-form-field>
+	<input matInput 
+		[disabled]="isDisabled"
+		[matDatepicker]="picker"
```
SuperITMan: What happens if we want to use multiple `stark-date-picker` components on the same page ?
I think that the id of the date pickers will be the same and that means no date-picker at all will work. 

As it was the case in the old Stark, I think we should provide a binding for setting the name/id of the date-picker to allow to use multiple ones in the same HTML page

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
```diff
@@ -0,0 +1,14 @@
...
+		[disabled]="isDisabled"
+		[matDatepicker]="picker" 
+		[placeholder]="label | translate" 
+		[matDatepickerFilter]="dateFilterFnWrapper"
+		[min]="minDate"
+		[max]="maxDate"
+		[name]="pickerName"
+		[value]="date"
+		(dateChange)="onDateChange($event)">
+	<mat-datepicker-toggle matSuffix [for]="picker"></mat-datepicker-toggle>
```
SuperITMan: Same here about the id

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
```diff
@@ -0,0 +1,14 @@
...
+	<mat-datepicker #picker></mat-datepicker>
```
SuperITMan: Same here about the id

catlabs:
@SuperITMan true, I can change the parse format but dynamic change of this format is not supported yet :-(
If we want to keep the current implementation, we need to choose between DD/MM/YYYY and MM/DD/YYYY.
https://github.com/angular/material2/issues/8355

Other solution will be to implement our own provider for date but it is not obvious and it will be time consuming.

What do we do?

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
christophercr: I think there is no way to set an id to this toggle element

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
catlabs: Well true we had an id before. 
The reason why is that we won't be able to use any functions on the mat-date-picker getting it by id. Instead of that, the StarkDatePicker expose a public picker: MatDatepicker and a public pickerInput: MatDatepickerInput, so the end user have access to them in the TS and use the API provided by Material.
https://material.angular.io/components/datepicker/api

christophercr:
@catlabs @SuperITMan I don't think that dynamic format should be a real issue for now, we can just specify that there would be only one format and cannot be changed. We can create an issue to track this and see if the Angular Material team implements that feature somewhere soon.

And for the formats, as @catlabs mentioned, we should just pick one format for the showcase and tell the user which format is being used (maybe in the placeholder of the picker) so it is clear how the example is supposed to work.

What do you think?

catlabs:
Also the end user can provide his own settings so they can decide what they want

```
export const STARK_DATE_FORMATS: Object = {
	parse: {
		dateInput: "LL"
	},
	display: {
		dateInput: "LL",
		monthYearLabel: "MMM YYYY",
		dateA11yLabel: "LL",
		monthYearA11yLabel: "MMMM YYYY"
	}
};
```

packages/stark-ui/src/modules/date-picker/components/date-picker.component.html
christophercr: Yep but the issue is not about interacting with the date-picker programatically, it is more about 2 other aspects as far as I can remember:
- the default HTML behaviour. In case the datepicker is inside a `<form>`, whenever such form is submitted, the fields of the form will be sent with the name of the `name` attribute of the `<input>` element. I think the same will be important in case we link an Angular Form Control with the datepicker.
- for the e2e testing, it would be easier to select a datepicker based on its `id` or `name` rather than with the CSS selector which might change more often than the attributes.

catlabs: The input name is already set:
```
@Input()
public pickerName: string = "";
```

About the id, true that it will be useful for e2e :-)

catlabs:
@christophercr @SuperITMan it's repushed

showcase/src/app/app.component.html
```diff
@@ -34,6 +34,9 @@
 			<a mat-list-item uiSref="demo-button" uiSrefActive="active">
 				<span matLine>Button</span>
 			</a>
+			<a mat-list-item uiSref="demo-date-picker" uiSrefActive="active">
+					<span matLine>Date Picker</span>
```
SuperITMan: There is an alignment issue. Could you please solve it ?

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -0,0 +1,139 @@
...
+	templateUrl: "./date-picker.component.html",
+	encapsulation: ViewEncapsulation.None
+})
+export class StarkDatePickerComponent {
+	/**
+	 * id attribute of the form field wrapping the mat-datepicker
+	 * id attribute followed by "-input" of the mat-datepicker-input
+	 *
+	 */
+	@Input()
```
SuperITMan: It's difficult to understand the inputs available in the code because it is not sorted in alphabetic order.
Could you please sort your inputs / outputs (based on their name) ? 😊 (I think that will be needed also to generate the documentation correctly.

showcase/src/assets/examples/date-picker/weekends.html
```diff
@@ -0,0 +1,10 @@
+<stark-date-picker 
+    pickerId="picker-only-weekends"
+    pickerName="only-weekends"
+    [date]="currentDate" 
+    [dateFilter]="'OnlyWeekends'" 
+    label="Choisir une date"
```
SuperITMan: I think all our example texts should be in English. What do you think about this ?

showcase/src/assets/examples/date-picker/weekdays.html
```diff
@@ -0,0 +1,10 @@
+<stark-date-picker 
+    pickerId="picker-only-weekdays"
+    name="only-weekdays"
+    [date]="currentDate" 
+    [dateFilter]="'OnlyWeekdays'" 
+    label="Choisir une date"
```
SuperITMan: I think all our example texts should be in English. What do you think about this ?

showcase/src/app/app.component.html
catlabs: Indeed, I did run prettier-check though...

catlabs:
Repushed

Issue #575, opened 2018-07-31
showcase: ExampleViewer does not take into account the baseUrl while fetching the example files


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

The ExampleViewer does not show any example files due to an Http error while fetching the files:

```
GET https://stark.nbb.be/assets/examples/keyboard-directives/on-enter-key-directive.html 404 ()
```

The problem occurs because the Http request triggered by the FileService to fetch the files does not take into account the baseUrl of the app when this is deployed on GitHub Pages.

## Expected behavior

The FileService should fetch the files targetting the correct Http, for example in the latest Showcase:

```
GET https://stark.nbb.be/showcase/latest/assets/examples/keyboard-directives/on-enter-key-directive.html 
```

## Minimal reproduction of the problem with instructions

Go on [Stark showcase](https://stark.nbb.be/showcase/latest/) and go to any demo page.
In the demo page click on the "View source" button of the ExampleViewer component.
The example files will not be displayed and an error will be shown in the console.

## Environment

<pre><code>
Angular version: X.Y.Z
Stark version: 10.0.0-alpha.4

Issue #580, opened 2018-07-31
showcase: Remove the tooltip from the ExampleViewer component


RobbyDeLaet:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When hovering over the example viewer, it displays the translation key from the title as a tool-tip.
Even when the mouse-pointer is over another component inside the example viewer.

## Expected behavior

Prevent the tool-tip from showing.

christophercr:
Closed by #584

PR #584, opened 2018-08-01
fix(stark-demo): adapt ExampleViewer to fetch example files targeting the right url. Rename title input to prevent weird tooltip.

christophercr:
ISSUES CLOSED: #575 #580

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #575 #580 

## What is the new behavior?
- The ExampleViewer fetches the example files from the correct url taking into account the base URL of the app when deployed in GitHub Pages.
- The ExampleViewer does not show a tooltip anymore when hovering over the component.

```
[ ] Yes
[X] No
```

### DIFF
--- a/showcase/src/app/demo/pretty-print/pretty-print.component.html
+++ b/showcase/src/app/demo/pretty-print/pretty-print.component.html
@@ -1,7 +1,7 @@
 <example-viewer *ngFor="let fileType of fileTypes; trackBy: trackFileTypes"
 				[extensions]="['HTML', 'SCSS', 'TS']"
 				[filesPath]="fileType.path"
-				[title]="fileType.title">
+				[exampleTitle]="fileType.title">
 	<mat-tab-group>
 		<mat-tab label="{{ 'SHOWCASE.DEMO.PRETTY-PRINT.ORIGINAL_CODE'  | translate }}">
 			<div class="code-block">{{ fileType.rawData }}</div>
--- a/showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts
+++ b/showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts
@@ -1,31 +1,34 @@
-import { HttpClientModule } from "@angular/common/http";
-import { BrowserAnimationsModule } from "@angular/platform-browser/animations";
-import { async, ComponentFixtureAutoDetect, ComponentFixture, TestBed } from "@angular/core/testing";
-import { MatButtonModule, MatTabsModule, MatTooltipModule } from "@angular/material";
-
-import { ExampleViewerComponent } from "./example-viewer.component";
-import { FileService } from "./file.service";
-import { throwError, of } from "rxjs";
-import { MockStarkLoggingService } from "@nationalbankbelgium/stark-core/testing";
-import { STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
-import SpyObj = jasmine.SpyObj;
-import { StarkPrettyPrintModule } from "@nationalbankbelgium/stark-ui";
 import { NO_ERRORS_SCHEMA } from "@angular/core";
+import { NoopAnimationsModule } from "@angular/platform-browser/animations";
+import { async, ComponentFixture, TestBed } from "@angular/core/testing";
+import { MatButtonModule, MatTabsModule, MatTooltipModule } from "@angular/material";
+import { throwError, of, Observable, Subject } from "rxjs";
+import { filter, delay } from "rxjs/operators";
+import { STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
+import { MockStarkLoggingService } from "@nationalbankbelgium/stark-core/testing";
+import { StarkPrettyPrintModule } from "@nationalbankbelgium/stark-ui";
+
+import { ExampleFile, ExampleViewerComponent } from "./example-viewer.component";
+import { FileService } from "./file.service";
+import SpyObj = jasmine.SpyObj;
+import Spy = jasmine.Spy;
 
 describe("ExampleViewerComponent", () => {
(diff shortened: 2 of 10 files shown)


christophercr:
@SuperITMan @catlabs I've fixed the issues with the failing tests of this PR. Could you please review it?

showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts
```diff
@@ -1,80 +1,147 @@
...
 
 	beforeEach(() => {
 		logger = TestBed.get(STARK_LOGGING_SERVICE);
+		fileService = TestBed.get(FileService);
+		fileService.fetchFile.and.callFake(() => of("initial dummy file content"));
+
 		fixture = TestBed.createComponent(ExampleViewerComponent);
 		component = fixture.componentInstance;
-		fileService = fixture.debugElement.injector.get<FileService>(FileService as any);
+		fixture.detectChanges(); // trigger initial data binding
```
catlabs: Don't you think it will be easier to use ComponentFixtureAutoDetect

https://angular.io/guide/testing#automatic-change-detection

showcase/src/app/shared/example-viewer/example-viewer.component.ts
```diff
@@ -3,56 +3,92 @@ import { HttpErrorResponse } from "@angular/common/http";
...
 	public ngOnInit(): void {
-		this.fetchFiles();
+		this.fetchExampleFiles();
 	}
 
-	public fetchFiles(): void {
-		this.extensions.forEach((extension: string) => {
-			this.service.fetchFile("/assets/examples/" + this.filesPath + "." + extension.toLowerCase()).subscribe(
+	public fetchExampleFiles(): void {
+		for (const extension of this.extensions) {
```
catlabs: Is there a reason here to use for over forEach?

showcase/src/app/demo/pretty-print/pretty-print.component.html
```diff
@@ -1,7 +1,7 @@
 <example-viewer *ngFor="let fileType of fileTypes; trackBy: trackFileTypes"
 				[extensions]="['HTML', 'SCSS', 'TS']"
 				[filesPath]="fileType.path"
-				[title]="fileType.title">
+				exampleTitle="fileType.title">
```
catlabs: This is not working, it should be [exampleTitle]="fileType.title"

showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts
christophercr: I thought so, but then I realized it is not working in all cases. In the link you mention it is stated the following:

> The second and third test reveal an important limitation. The Angular testing environment does not know that the test changed the component's title. The ComponentFixtureAutoDetect service responds to asynchronous activities such as promise resolution, timers, and DOM events. But a direct, synchronous update of the component property is invisible. The test must call fixture.detectChanges() manually to trigger another cycle of change detection.

> Rather than wonder when the test fixture will or won't perform change detection, the samples in this guide always call detectChanges() explicitly. There is no harm in calling detectChanges() more often than is strictly necessary.

And to be honest I have the same feeling, I don't like the fact that it will work in some cases and it won't in some others (like in the line 45 of the old code in this test file).
Therefore I prefer to explicitly trigger change detection and know what is going on at all times and avoid some magic done by Angular in "some cases".

showcase/src/app/shared/example-viewer/example-viewer.component.ts
christophercr: Good question. Indeed, basically 2 reasons:

- Better performance: when transpiling TS code to Javascript, an `Array.forEach()` method is kept as is because is ES5 standard whereas `for...of` loop is converted to a basic `for` loop (when the target ES version is ES5) which is more performant in NodeJS and also in most Javascript engines of web browsers. I know that maybe in this specific case this loop will not have more than 3 or 5 iterations but I prefer to always use the `for...or` loop to keep consistency all over the code base and avoid confusions over when to use one syntax or the other.
- Shorter syntax: a `for...of` is shorter than a `forEach` because you don't need to specify the type of the item being iterated, it is inferred by Typescript.

showcase/src/app/demo/pretty-print/pretty-print.component.html
christophercr: Ups, indeed

showcase/src/app/shared/example-viewer/example-viewer.component.ts
catlabs: Allright I will also use for...of then :-)

showcase/src/app/shared/example-viewer/example-viewer.component.spec.ts
```diff
@@ -1,80 +1,147 @@
...
 	});
 
-	describe("@Input() title", () => {
-		it("should change the title according to the @Input", () => {
-			const h2: any = fixture.nativeElement.querySelector(".example-viewer-title-spacer h2");
-			component.title = "Test title";
+	describe("@Input() exampleTitle", () => {
+		it("should change the exampleTitle according to the @Input", () => {
+			const h2: HTMLHeadingElement = fixture.nativeElement.querySelector(".example-viewer-title-spacer h2");
+			component.exampleTitle = "Test exampleTitle";
```
SuperITMan: `"Test exampleTitle"` does not really make sense. Could you please put back `"Test title"` which has more sense ? 😊

christophercr: Ah true, the "find-replace" did a bit too much :p  I'll fix it

christophercr:
@SuperITMan I updated the PR with your remark. Can you review it please?

Issue #662, opened 2018-09-06
ui: SCSS - create global color settings


catlabs:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior
We should create global settings for colors that are added to primary, accent and warning so we can use those ones in different components if needed.
Here is for instance the extra button's colors:

```
$button-theme: (
  alert: (
    color: #d32f2f,
    contrast: $light-primary-text
  ),
  alt: (
    color: #323232,
    contrast: $light-primary-text
  ),
  neutral: (
    color: #bfbfbf,
    contrast: $light-primary-text
  ),
  success: (
    color: #4caf50,
    contrast: $light-primary-text
  )
);
```

## Expected behavior
We should rename this SCSS map and add all the extra colors in it.

catlabs:
We also need a strategy on how to customize colors on components. Following Material, they use `color="primary"` on buttons for instance. I think we should keep this for consistency. From there, we have 2 options:
- Read the color attribute and add the right classes to the component (Material way). Backward is that we might need to add extra code in our ts
- Read the color and add the color to subcomponents (ex: dropdown inside language selector). Backward is that we use attribute selector in SCSS

PR #672, opened 2018-09-10
feat(stark-ui): typography + Spaces

catlabs:
ISSUES CLOSED: #475, #671

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #475, #671

## What is the new behavior?
- Typography: h1, h2, h3, h4 matching with Material font map. I also changed the pretty-print font to use the same as the rest.
- Spaces: a `<section>` tag helps to organize the layout structure.

Check out the typography page, if we agree on this layout, I will then integrate it on the other pages.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #678, opened 2018-09-11
feat(stark-ui): color settings

catlabs:
ISSUES CLOSED: #662

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #662

## What is the new behavior?
The end user has now a global simplified SCSS map to set the colors but he can still uses the full palette.
Add an input to the dropdown so its color can be set by a parent component (LanguageSelector for instance)

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/stark-ui/assets/styles/_colors.scss
+++ b/packages/stark-ui/assets/styles/_colors.scss
@@ -13,9 +13,9 @@ $mat-light-theme-foreground: $foregrounds;
   }
 }
 
-$stark-color-theme: mat-light-theme(
+$theme: mat-light-theme(
   map-get($base-theme, primary-palette),
   map-get($base-theme, accent-palette),
   map-get($base-theme, warn-palette)
 ) !default;
-@include angular-material-theme($stark-color-theme);
+@include angular-material-theme($theme);
--- a/packages/stark-ui/assets/styles/_variables.scss
+++ b/packages/stark-ui/assets/styles/_variables.scss
@@ -1,4 +1,41 @@
-$border-radius: 2px;
+$color-theme: (
+  accent: (
+    color: #8f0039,
+    contrast: $light-primary-text
+  ),
+  alert: (
+    color: #d32f2f,
+    contrast: $light-primary-text
+  ),
+  alt: (
+    color: #323232,
+    contrast: $light-primary-text
+  ),
+  neutral: (
+    color: #bfbfbf,
+    contrast: $light-primary-text
+  ),
+  primary: (
+    color: #0076c8,
+    contrast: $light-primary-text
+  ),
+  success: (
+    color: #4caf50,
+    contrast: $light-primary-text
+  ),
+  warning: (
+    color: #ff9800,
+    contrast: $light-primary-text
+  ),
+  white-transp: (
+    color: rgba(255, 255, 255, 0.87),
+    contrast: $dark-primary-text
+  )
+);
+
+@if variable-exists(stark-color-theme) {
+  $color-theme: map-merge($color-theme, $stark-color-theme);
+}
 
 $primary-palette: (
   50: #e3f2fd,
@@ -6,7 +43,7 @@ $primary-palette: (
   200: #90caf9,
   300: #64b5f6,
   400: #42a5f5,
-  500: #0076c8,
+  500: map-get(map-get($color-theme, primary), color),
   600: #006ab4,
   700: #005ea0,
   800: #1565c0,
@@ -38,7 +75,7 @@ $accent-palette: (
   200: #f48fb1,
   300: #f06292,
   400: #ec407a,
-  500: #8f0039,
+  500: map-get(map-get($color-theme, accent), color),
   600: #70002d,
   700: #c2185b,
   800: #ad1457,
@@ -70,7 +107,7 @@ $warning-palette: (
   200: #ffcc80,
   300: #ffb74d,
   400: #ffa726,
-  500: #ff9800,
+  500: map-get(map-get($color-theme, warning), color),
   600: #fb8c00,
   700: #f57c00,
   800: #ef6c00,
@@ -132,35 +169,6 @@ $base-theme: (
   primary-palette: mat-palette($primary-palette),
   accent-palette: mat-palette($accent-palette),
   warn-palette: mat-palette($warning-palette)
-);
-$button-theme: (
-  alert: (
-    color: #d32f2f,
-    contrast: $light-primary-text
-  ),
-  alt: (
-    color: #323232,
-    contrast: $light-primary-text
-  ),
-  neutral: (
-    color: #bfbfbf,
-    contrast: $light-primary-text
-  ),
-  success: (
-    color: #4caf50,
-    contrast: $light-primary-text
-  )
-);
-$dropdown-theme: (
-  white-transp: (
-    color: rgba(255, 255, 255, 0.87)
-  )
-);
-$card-theme: (
-  border-radius: $border-radius
-);
-$table-theme: (
-  border-radius: $border-radius
 );
 
 // font-size, line-height, font-weight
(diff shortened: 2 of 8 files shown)


packages/stark-ui/assets/styles/_colors.scss
```diff
@@ -13,9 +13,9 @@ $mat-light-theme-foreground: $foregrounds;
   }
 }
 
-$stark-color-theme: mat-light-theme(
+$theme: mat-light-theme(
```
christophercr: Shouldn't be the name of this "stark-theme" instead of just "theme". That way it is clear it is something from "Stark"..

packages/stark-ui/assets/styles/_variables.scss
```diff
@@ -1,12 +1,49 @@
...
+    contrast: $light-primary-text
+  ),
+  white-transp: (
+    color: rgba(255, 255, 255, 0.87),
+    contrast: $dark-primary-text
+  )
+);
+
+@if variable-exists(stark-color-theme) {
+  $color-theme: map-merge($color-theme, $stark-color-theme);
```
christophercr: "$stark-color-theme" ? Wasn't it renamed in the file above?

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -123,6 +140,10 @@ export class StarkDropdownComponent implements OnInit, OnChanges, OnInit {
 		this.logger.debug(componentName + ": component initialized");
 		this.optionsAreSimpleTypes = this.areSimpleTypes();
 		this.setDefaultBlank();
+		if (this.color) {
+			const nativeElement: HTMLElement = this.elementRef.nativeElement;
+			nativeElement.classList.add("stark-" + this.color);
```
christophercr: We should try to use the Angular's [Renderer2](https://angular.io/api/core/Renderer2#addClass) for this instead of directly manipulating the DOM ;)

christophercr: I don't think we should add the `stark-`prefix to the color, otherwise in case the developer wants to use his own color (i.e.   red), he should add a CSS rule for the `stark-red` but in fact it is not an stark color.... why not simply calling the colors like `color-white`, `color-red` and so on? Or even better to name them `color-theme-white`, `color-theme-red` so it is clear that those are linked to the `$color-theme` map. What do you think?

packages/stark-ui/assets/styles/_variables.scss
```diff
@@ -1,12 +1,49 @@
...
+}
 
 $primary-palette: (
   50: #e3f2fd,
```
christophercr: Shouldn't all the shades be also customizable? Otherwise it might be that the color will be completely different when using shades 500 and any other 😕

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
christophercr: Thinking a bit more on this, couldn't we abstract this new property in an abstract class that we could extend from? This way it will be easier to apply the same functionality in other stark components...

packages/stark-ui/assets/styles/_colors.scss
catlabs: All the `$stark-something` are maps that the end user can defined. On our side we declare a map call `$something` which will be merged with `$stark-something`. I think we should keep this convention.
In this example $theme is only a local var that we use to instanciate the theming, it won't be used anywhere else so the naming is not really important.

packages/stark-ui/assets/styles/_variables.scss
catlabs: That's indeed why I renamed above. We give the ability to provide a new map called $stark-color-theme, which contains a simple version of the colors (instead of the full palette). Well as soon as https://github.com/NationalBankBelgium/stark/pull/672 is merged, I will add the doc for that (https://github.com/NationalBankBelgium/stark/issues/679). 

Robby need this one to be merged to finish the language selector, that's why I have separate the doc :-)

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
catlabs: About the first remark, here is our new stark color map:

```
$color-theme: (
  accent: (
    color: #8f0039,
    contrast: $light-primary-text
  ),
  alert: (
    color: #d32f2f,
    contrast: $light-primary-text
  ),
  alt: (
    color: #323232,
    contrast: $light-primary-text
  ),
  neutral: (
    color: #bfbfbf,
    contrast: $light-primary-text
  ),
  primary: (
    color: #0076c8,
    contrast: $light-primary-text
  ),
  success: (
    color: #4caf50,
    contrast: $light-primary-text
  ),
  warning: (
    color: #ff9800,
    contrast: $light-primary-text
  ),
  white-transp: (
    color: rgba(255, 255, 255, 0.87),
    contrast: $dark-primary-text
  )
);
```
And the idea was to call the color stark-accent, stark-primary etc... so to me it is quite clear this way. If you check the `_base.css` in stark-ui, all the stark class have stark as a prefix, so to me it make more sense keeping it this way for every provided classes.

@SuperITMan any opinion on this?

packages/stark-ui/assets/styles/_variables.scss
catlabs: yes they will have the choice, either to use the simple array, either to override the full palettes. It can be heavy to ask them to override all the palettes, so I think we should keep a simple version of the map. Of course we can add new color as we go on.

catlabs:
Repushed

packages/stark-ui/src/common/classes/abstract-color.ts
```diff
@@ -0,0 +1,23 @@
+import { OnInit, Input, Renderer2, ElementRef } from "@angular/core";
+
+/**
+ * Abstract class to add the right stark class color to the elementRef
+ */
+export abstract class StarkAbstractColor implements OnInit {
```
christophercr: Could we rename this class to `AbstractStarkUiComponent`? Discussing a bit more with @SuperITMan it is a good idea that all components in Stark UI extend from it since all those components might need to be themable.

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -19,7 +32,13 @@ const componentName: string = "stark-dropdown";
 		class: componentName
 	}
 })
-export class StarkDropdownComponent implements OnInit, OnChanges, OnInit {
+export class StarkDropdownComponent extends StarkAbstractColor implements OnInit, OnChanges, OnInit {
+	/**
+	 * Color theme of the dropdown
+	 */
+	@Input()
+	public color?: string;
```
christophercr: There is no need to re-declare the `color` property here, otherwise there is no benefit of having an abstract ;)

packages/stark-ui/src/common/classes/abstract-color.ts
```diff
@@ -0,0 +1,23 @@
...
+	/**
+	 * Color theme
+	 */
+	@Input()
+	public color?: string;
```
christophercr: Let's make this property `protected`

packages/stark-ui/src/common/classes/abstract-color.ts
```diff
@@ -0,0 +1,23 @@
...
+
+	public constructor(protected renderer: Renderer2, protected elementRef: ElementRef) {}
```
christophercr: Let's make this constructor `protected`

catlabs:
Rebased and repushed :-)

Issue #523, opened 2018-07-17
ui: table - add support for customizing styles (header, column, row and cell)


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Expected behavior

The developer should be able to define custom styling for an specific parts of the stark-table:

- header
- column
- row
- cell

PR #857, opened 2018-11-14
feat(stark-ui): add custom styling functionality to stark table component

cnomes:
ISSUES CLOSED: #523

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Cannot add classes to table cells / headers / rows / columns

Issue Number: #523 

## What is the new behavior?
Classes can be added through properties on the table component and styling added to those classes via SCSS.

- added properties to stark-table component to allow for custom styling
- added tests for new functionality
- added demo component demonstrating new functionality

```
[ ] Yes
[X] No
```

## Other information


packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -139,6 +139,12 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	@Input()
 	public tableRowsActionBarConfig: StarkActionBarConfig;
 
+	/**
+	 * function to generate a classNames for rows
```
christophercr: Typo: "generate a classNames" -> "generate classNames"

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -224,4 +246,36 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent {
...
+		const classNameFnResult: string =
+			(this.cellClassNameFn && typeof this.cellClassNameFn === "function" && this.cellClassNameFn(value, row, this.name)) || "";
+		return `${classNameFnResult} ${this.cellClassName || ""}`;
+	}
+
+	/**
+	 * Get the classes for a header
+	 * @returns The classes for the header
+	 */
+	public getHeaderClassName(): string {
```
christophercr: Isn't it better to call it "getHeaderClassNames" since it can return multiple classes?

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -224,4 +246,36 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent {
 	public onSortChange(): void {
 		this.sortChanged.emit(this);
 	}
+
+	/**
+	 * Gets a the classes for a specific cell, based on the cellClassName Input and the cellClassNameFn function if it was given.
+	 * @param row - The data object of the row the cell is in.
+	 * @returns The classes for the cell.
+	 */
+	public getCellClassName(row: any): string {
```
christophercr: Isn't it better to call it "getCellClassNames" since it can return multiple classes?

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -714,6 +724,16 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 		return filterValueReset;
 	}
 
+	/**
+	 * Gets the class for a specific row if a rowClassNameFn function has been given as an Input
+	 * @param row - The data object passed to the row.
+	 * @param index - The index of the row.
+	 * @returns The className generated by the rowClassNameFn function
+	 */
+	public getRowClass(row: any, index: number): string {
```
christophercr: Isn't it better to call it "getRowClasses" since it can return multiple classes?

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -714,6 +724,16 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
+		return this.rowClassNameFn && typeof this.rowClassNameFn === "function" ? this.rowClassNameFn(row, index) : "";
```
christophercr: I think the first check can be removed... `this.rowClassNameFn && typeof this.rowClassNameFn === "function"``could be simply `typeof this.rowClassNameFn === "function"``;)

showcase/src/app/demo/table/demo-table.component.ts
```diff
@@ -32,8 +32,14 @@ export class DemoTableComponent implements OnInit {
 		return 0;
 	};
 
+	public getRowClassName = (_row: any, index: number): Object => (index % 2 === 0 ? "even" : "odd");
```
christophercr: We should use the `object` type instead of `Object`. Why? https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

showcase/src/app/demo/table/demo-table.component.ts
```diff
@@ -89,23 +95,30 @@ export class DemoTableComponent implements OnInit {
...
+				isFilterable: true,
+				compareFn: this.compareTitle,
+				cellClassNameFn: this.getTitleCellClassName
 			},
 			{
 				name: "description",
-				sortable: true
+				label: "Description",
+				isSortable: true,
+				isFilterable: true
```
christophercr: Can you also set a custom class on a row (`className` property) to demo this feature?

showcase/src/assets/examples/table/with-custom-styling.scss
```diff
@@ -0,0 +1,19 @@
+.even td {
+  background-color: #d1e1ff;
```
christophercr: Regarding all the styles in this SCSS file, I suggest we remove the element selectors like `td` and `th`... otherwise it feels like we are cheating setting directly the styles for specific HTML tags 😆 which might not always be the case if we change our Table component at some point

showcase/src/app/demo/table/demo-table.component.scss
```diff
@@ -18,3 +18,39 @@
     margin-bottom: 0;
   }
 }
+
+.even td {
+  background-color: #d1e1ff;
```
christophercr: Regarding all the styles in this SCSS file, I suggest we remove the element selectors like `td` and `th`... otherwise it feels like we are cheating setting directly the styles for specific HTML tags 😆 which might not always be the case if we change our Table component at some point

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -893,4 +897,51 @@ describe("TableComponent", () => {
 			expect((<StarkTableColumnFilter[]>component.filter.columns)[0].filterValue).toBe(dummyColumnFilterValue);
 		});
 	});
+
+	describe("setStyling", () => {
+		const dummyData: Array<any> = [
```
christophercr: Same here: could you use the `[]` syntax for arrays so we keep consistency in our codebase?

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -893,4 +897,51 @@ describe("TableComponent", () => {
...
+			hostComponent.dummyData = dummyData;
+
+			hostFixture.detectChanges(); // trigger data binding
+			component.ngAfterViewInit();
+		});
+
+		it("second row should have class 'even'", () => {
+			const tableElement: HTMLElement = hostFixture.nativeElement;
+			const secondRow: HTMLElement | null = tableElement.querySelectorAll("tr").item(1);
+			expect(secondRow && secondRow.classList).toContain("even");
```
christophercr: Maybe you can check in this test also that the odd row has no class set ;)

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -893,4 +897,51 @@ describe("TableComponent", () => {
...
+		});
+
+		it(`cell description should have class 'staticCellClass'`, () => {
+			const tableElement: HTMLElement = hostFixture.nativeElement;
+			const descriptionCell: HTMLElement | null = tableElement.querySelector("tr td:nth-child(2)"); // select the description cells
+			expect(descriptionCell && descriptionCell.classList).toContain(staticCellClass);
```
christophercr: Maybe you can check that the class is present in all the cells of the column

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -224,4 +246,36 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent {
...
+		const value: any | undefined = this.getRawValue(row);
+		const classNameFnResult: string =
+			(this.cellClassNameFn && typeof this.cellClassNameFn === "function" && this.cellClassNameFn(value, row, this.name)) || "";
+		return `${classNameFnResult} ${this.cellClassName || ""}`;
```
christophercr: Can you write a test for this code? I'm not sure whether this will output what we expect....

packages/stark-ui/src/modules/table/components/table.component.ts
SuperITMan: There is another small typo: **F**unction to...

packages/stark-ui/src/modules/table/components/table.component.ts
SuperITMan: I thought that we could only do `if (this.rowClassNameFn instanceof Function)` if we want to check if it is a function...
Which way should we use so ? :blush:
I've the feeling that `typeof this.rowClassNameFn === "function"` is better.

christophercr: true, that is what I am also suggesting... but the formatting of my comment was wrong :( I changed it

christophercr: True, but I suggest to do it in a different PR where we can refactor all the code not only for this but for other things like `readonly`, `private`, `get`, `set`, etc...

showcase/src/app/demo/table/demo-table.component.ts
cnomes: Even worse, it should be string. 😆

showcase/src/assets/examples/table/with-custom-styling.scss
```diff
@@ -0,0 +1,19 @@
...
+.warning {
+  color: #ff9800;
+}
+
+.success {
+  color: #3c9f40;
+}
+
+.bold {
+  font-weight: bold;
```
christophercr: This style is only applied to the Header of the column but not to all its cells 😞 

table-column-styles-issue

showcase/src/assets/examples/table/with-custom-styling.scss
```diff
@@ -0,0 +1,19 @@
+.even {
+  background-color: #d1e1ff;
```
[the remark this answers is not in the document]
cnomes: Ah, yes. 🙈 the `td` was there because it applies its own `background-color`

cnomes:
@christophercr I updated the PR. 😄

Issue #856, opened 2018-11-14
core: http - Error when performing a request with loggingFlushDisabled set to true


erwan-carlig:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When deserializing the `stark-app-config.json` (app.module.ts:99), by default the "loggingCorrelationIdHttpHeaderName" property is either set to the content of `stark-app-config.json` or to "correlation-id" (it can never be undefined).

If logging is disabled (`loggingFlushDisabled` set to true), the StarkLoggingServiceImpl will not generate a correlationId ([logging.service.ts:69](https://github.com/NationalBankBelgium/stark/blob/10.0.0-beta.1/packages/stark-core/src/modules/logging/services/logging.service.ts#L69)).

When using the stark http service, it will try to add the correlation headers ([http.service.ts:176](https://github.com/NationalBankBelgium/stark/blob/10.0.0-beta.1/packages/stark-core/src/modules/http/services/http.service.ts#L176)) and set the correlation id header to the value of correlationId coming from the StarkLoggingService (which is undefined since it was not generated).

When trying to perform the request it will cause any of the following errors: 
```
ERROR TypeError: Cannot read property 'keys' of undefined
    at StarkHttpServiceImpl.ac_main../node_modules/@nationalbankbelgium/stark-core/fesm5/stark-core.js.StarkHttpServiceImpl.getResponseHeaders 
```

```
"Cannot read property 'length' of undefined" in file headers.ts:74 (@angular/common/http/src/headers.ts)
```

## Expected behavior
No correlation header should be added if logging is disabled.

## Others:
Not a pressing issue, I can work around it.

## Environment

<pre><code>
Angular version: 6.1.10
Stark version: 10.0.0-beta.1

</code></pre>

christophercr:
Thanks for the report! We'll fix it in the next release ;)

erwan-carlig:
Hello,

Unfortunately the "loggingCorrelationIdHttpHeaderName" is added by the StarkHttpService's method "addCorrelationIdentifierHeader", which directly access the headers map and add the undefined header.
[...]

PR #881, opened 2018-11-23
fix(stark-core): refactor Http headers related API to prevent undefined and null header values to be added to Angular Http headers

christophercr:
ISSUES CLOSED: #856

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #856

## What is the new behavior?
The correlationId Http header is not added in case the correlation id is not defined (which is the case when the `loggingFlushDisabled` is true).

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-core/src/modules/http/builder/http-abstract-base-request-builder.ts
+++ b/packages/stark-core/src/modules/http/builder/http-abstract-base-request-builder.ts
@@ -12,8 +12,9 @@ export abstract class StarkAbstractHttpBaseRequestBuilder<T extends StarkResourc
 		this.request = request;
 	}
 
-	public setHeader(name: string, value: string): this {
-		if (typeof value !== "undefined") {
+	public setHeader(name: string, value: string | string[]): this {
+		// in Angular, a header value can only be string or string[], not null/undefined (https://github.com/angular/angular/issues/18743)
+		if (name && typeof value !== "undefined" && value !== null) {
 			this.request.headers.set(name, value);
 		}
 		return this;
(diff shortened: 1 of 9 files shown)


packages/stark-core/src/modules/http/builder/http-abstract-base-request-builder.ts
```diff
@@ -12,8 +12,9 @@ export abstract class StarkAbstractHttpBaseRequestBuilder<T extends StarkResourc
 		this.request = request;
 	}
 
-	public setHeader(name: string, value: string): this {
-		if (typeof value !== "undefined") {
+	public setHeader(name: string, value: string | string[]): this {
+		// in Angular, a header value can only be string or string[], not null/undefined (https://github.com/angular/angular/issues/18743)
+		if (typeof value !== "undefined" && value !== null) {
```
SuperITMan: I understand that we cannot set undefined or null as header value but we don't check the name itself.
What would happen if the name was undefined or null ?

packages/stark-core/src/modules/session/services/session.service.ts
```diff
@@ -358,7 +358,12 @@ export class StarkSessionServiceImpl implements StarkSessionService {
 			this._devAuthenticationHeaders = new Map<string, string>();
 		}
 
-		devAuthenticationHeaders.forEach((value: string, key: string) => this._devAuthenticationHeaders.set(key, value));
+		devAuthenticationHeaders.forEach((value: string, key: string) => {
```
SuperITMan: Should not we change this to `(value: string | string[], key: string) => {`?

Because you mention in the comment above that Angular supports both but it is not implemented in this case...

christophercr: Indeed, I'll fix this

packages/stark-core/src/modules/http/builder/http-abstract-base-request-builder.ts
christophercr: True, I'll fix this

Issue #859, opened 2018-11-15
showcase: 404 errors on github pages


dsebastien:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior
Going to directly to the URL of a specific route in the showcase application hosted on github returns a 404 error. Here's an example: https://stark.nbb.be/showcase/latest/demo/keyboard-directives

This is problematic because it means that we can't middle-click on specific pages or bookmark them.

## Expected behavior
Being able to access the page directly

dsebastien:
A solution is described here: https://github.com/rafrex/spa-github-pages
We need to add a 404.html page with a script to handle the redirection.

This is just a workaround; the real issue will remain in GH pages: https://github.com/isaacs/github/issues/408

Mallikki:
This was resolved thanks to the following link: http://www.backalleycoder.com/2016/05/13/sghpa-the-single-page-app-hack-for-github-pages/

christophercr:
@Mallikki I re-opened this issue because this fix should also be implemented in the Showcase and the Starter so we don't have to add this fix again on every release ;)

PR #891, opened 2018-11-27
fix(stark-showcase):add a 404.html page to fix the 404 error set by Github servers 

Mallikki:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```
Issue Number: #859 

## What is the new behavior?

No 404 error when we refresh a page anymore

```
[ ] Yes
[x] No
```


christophercr:
Looks good to me.

@dsebastien @SuperITMan in this case the solution is a bit different from what is proposed [here]( http://www.backalleycoder.com/2016/05/13/sghpa-the-single-page-app-hack-for-github-pages/): we use Javascript `location` to do the redirection which the article says it is **"arguably worse for crawler indexing"**. What do you think about this?

christophercr:
@Olivia, I don't think we need the "404" page in the Showcase. This should be present only at the root of the _gh-pages_ branch... so we only need this one: https://github.com/NationalBankBelgium/stark/blob/gh-pages/404.html

showcase/404.html
```diff
@@ -0,0 +1,22 @@
+<html>
+<head>
+	<script>
+		sessionStorage.redirect = location.href;
```
christophercr: @Olivia, I don't think we need the "404" page in the Showcase. This should be present only at the root of the _gh-pages_ branch... so we only need this one: https://github.com/NationalBankBelgium/stark/blob/gh-pages/404.html

dsebastien:
@christophercr Good point, this isn't great for SEO but we can live with that for now. Maybe create a new ticket for the improvement?

showcase/404.html
Mallikki: yup, makes sense actually --' thanks!

Issue #207, opened 2018-03-07
ui: route search component


dsebastien:
Integrate the "route search" component.

This component should have an input field and a search button, that provides auto completion for routes in the application. Once a route is selected, the app navigates to that route.

That component should have two display modes:
* [ ] one just like we have currently on the homepage of Cash
* [ ] one for displaying it in a more compact way in the header. It should display a Search icon by default and clicking on it should animate the display of the search field

christophercr:
@Mallikki these are some ideas for the search component 😉 

https://material.io/design/navigation/search.html#expandable-search
https://teradata.github.io/covalent/#/components/search

PR #915, opened 2018-12-05
feat(stark-ui): implementation of a route search component

Mallikki:
ISSUES CLOSED: #207

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```
#207 

## What is the new behavior?

A route search component has been added and allow the user to search for a specific page and navigate to it directly

```
[ ] Yes
[x] No
```

## Other information


Mallikki:
@christophercr I think we could improve the look of it a little bit! For example, maybe set the underline color a bit brighter? 
Also, the list of routes in the demo page is not the same as the one in the header, is it because they are not defined in the same folder?

packages/stark-ui/src/modules/route-search/components/route-search.component.html
```diff
@@ -0,0 +1,58 @@
+<div *ngIf="direction === 'left'">
```
christophercr: As we discussed, instead of duplicating the HTML just to put the search button to the left or to the right, we can do this only with css: [flex-direction: row-reverse](https://css-tricks.com/almanac/properties/f/flex-direction/)

packages/stark-ui/src/modules/route-search/components/route-search.component.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+	let component: RouteSearchComponent;
+	let hostComponent: TestHostComponent;
+	let hostFixture: ComponentFixture<TestHostComponent>;
+
+	const mockRoutingService: StarkRoutingService = new MockStarkRoutingService();
+
+	const optionsList: RouteEntry[] = [
+		{
+			path: "path 1",
+			url: "/url1"
```
christophercr: Weird that we can specify the "url" property on a `RouteEntry` because such interface doesn't have that property 😕

packages/stark-ui/src/modules/route-search/components/route-search.component.spec.ts
```diff
@@ -0,0 +1,153 @@
+/* tslint:disable:completed-docs no-commented-code prefer-const*/
```
christophercr: Why is the `prefer-const` rule disabled? This should never be disabled since it is really easy to fix 😉

packages/stark-ui/src/modules/route-search/components/route-search.component.spec.ts
Mallikki: Yup, I had to push my branch directly after having done a lot of changes that haven't been tested yet that's why everything is so messed up. I had to work on the app-data urgently so we can make a release by tuesday, I'll continue with the route search right after that ;)

Mallikki:
@christophercr @SuperITMan I've managed to retrieve all information we need from the menu =) I've commited and pushed so you can already have a look, but I still have a lot to do. I need to improve the template, the way we retrieve routes from UIRouter, reactivate the tests, etc... But it's working better than before and the routes look very good in the header.
The user will have a choice here: to pass his/her own menu or to use the one defined by UirRouter (in that case, he/she doesn't have to pass anything to the template).

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
+import {Component, Inject, Input, OnInit, ViewEncapsulation} from "@angular/core";
+import { RouteEntry } from "../entities/route-entry";
+import { Observable } from "rxjs/internal/Observable";
```
christophercr: Wrong import it should be: `import { Observable } from "rxjs";`

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+import { FormControl } from "@angular/forms";
+import {map, startWith} from "rxjs/operators";
+import { STARK_ROUTING_SERVICE, StarkRoutingService, STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
+import { UIRouter } from "@uirouter/core";
+import {StarkMenuConfig, StarkMenuGroup, StarkMenuSection} from "@nationalbankbelgium/stark-ui";
```
christophercr: Wrong import `@nationalbankbelgium/stark-ui`

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+export type StarkRouteSearchDirection = "left" | "right" | "";
```
christophercr: We should set a default direction instead of using `""`. Can you remove the empty string from these types?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+	
+	/**
+	 * the controller of the form
+	 */
+	public pathCtrl: FormControl;
+
+	/**
+	 * the list of routes filtered by autocomplete
+	 */
+	public filteredOptions: Observable<any[]>;
```
christophercr: We should specify the type of the expected value instead of just `any`

packages/stark-ui/src/modules/route-search/components/route-search.component.spec.ts
```diff
@@ -0,0 +1,274 @@
+/* tslint:disable:completed-docs no-commented-code*/
+import {MatTooltipModule} from "@angular/material/tooltip";
+import {async, ComponentFixture, TestBed} from "@angular/core/testing";
+import {RouteEntry, StarkMenuConfig, StarkMessageType} from "@nationalbankbelgium/stark-ui";
```
christophercr: Wrong import `@nationalbankbelgium/stark-ui`

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		const filterValue: string = value.toLowerCase();
+		return this.optionsToDisplay.filter((state: RouteEntry) => state.label.toLowerCase().indexOf(filterValue) === 0);
+	}
+
+	
+	public ngOnInit(): void {
+
+		this.inputValue="";
+		this.filteredOptions = this.pathCtrl.valueChanges.pipe(
+			startWith(""),
```
christophercr: Why is this initialized to an empty string? Isn't it better to use a `map` operator instead to convert an undefined value in `""`?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+			map((option: string) => (option ? this.filterOptions(option) : this.optionsToDisplay.slice()))
```
christophercr: Let's use ES6 instead... so you can replace `this.optionsToDisplay.slice()` by `[...this.optionsToDisplay]`  :wink:

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		this.sortPathList();
+		this.inputValue="";
+	}
+
+	/**
+	 * redirect the user to the chosen path
+	 */
+	public redirect(value: string): void {
+		this.inputValue="";
+		this.starkRoutingService.navigateTo(value);
```
christophercr: We also need to pass the state parameters here ...

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+	/**
+	 * the list of routes filtered by autocomplete
+	 */
+	public optionsToDisplay: RouteEntry[] = [];
+	
+	/**
+	 * The menu which options will be displayed
+	 */
+	@Input()
+	public optionsList?: StarkMenuConfig;
```
christophercr: Can you rename this variable to `menuConfig` so it's clear what content is expected to be there?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+
+	/**
+	 * the list of routes filtered by autocomplete
+	 */
+	public filteredOptions: Observable<any[]>;
+
+	/**
+	 * the list of routes filtered by autocomplete
+	 */
+	public optionsToDisplay: RouteEntry[] = [];
```
christophercr: Can you rename this variable to `routesToDisplay` so it is clear what is the content we should put in there?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+
+		if (this.optionsToDisplay !== undefined){
+			if (this.optionsList === undefined){
+				this.optionsToDisplay =	this.initRoutePathFromRouter();
+			} else{
+				this.optionsToDisplay = this.initRoutePathFromConfig();
+			}
+		}
+		this.sortPathList();
+		this.inputValue="";
```
christophercr: This line is duplicated... it is already the first line of this `ngOnInit`

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	/**
+	 * Retrieve the list of routes from a StarkMenuConfig object which the user have passed to the component
+	 */
+	//tslint:disable:cognitive-complexity
+	public initRoutePathFromConfig(): RouteEntry[]{
+		const optionsToDisplay: RouteEntry[] = [];
+		if (this.optionsList !== undefined && this.optionsList.menuSections !== undefined) {
+			this.optionsList.menuSections.forEach((section: StarkMenuSection) => {
```
christophercr: We should use `for ... of` loops instead of `Array.forEach()`. Basically for 2 reasons:

- Better performance: when transpiling TS code to Javascript, an `Array.forEach()` method is kept as is because is ES5 standard whereas `for...of` loop is converted to a basic `for` loop (when the target ES version is ES5) which is more performant in NodeJS and also in most Javascript engines of web browsers. I know that maybe in this specific case this loop will not have more than 3 or 5 iterations but I prefer to always use the `for...of` loop to keep consistency all over the code base and avoid confusions over when to use one syntax or the other.
- Shorter syntax: a `for...of` is shorter than a `forEach` because you don't need to specify the type of the item being iterated, it is inferred by Typescript.

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+				if (state.name !== undefined && state.url !== undefined && state.url.length > 1) {
+					let stateUrl: string = "";
+					if (state.url.startsWith("^/")) {
+						stateUrl = state.url.substring(2, state.url.length);
+					} else if (state.url.startsWith("/")) {
+						stateUrl = state.url.substring(1, state.url.length);
+					} else {
+						stateUrl = state.url.toString();
+					}
+					optionsToDisplay.push({ label: state.name.toString(), targetState: stateUrl });
```
christophercr: I think this is not correct... we are passing the URL as the `targetState`... but in fact we should pass the target state which is `state` otherwise it feels like we are circumventing our own API :p

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		return optionsToDisplay;
+	}
+
+	/**
+	 * Retrieve the list of routes from UIRouter if the user haven't passed a list to the component
+	 */
+	public initRoutePathFromRouter(): RouteEntry[]{
+		const optionsToDisplay: RouteEntry[] = [];
+		if (this.optionsList === undefined){
+			for (const state of this.routingService.stateService.get()) {
```
christophercr: For this logic, we should also think about the lazy loaded modules to avoid loading the modules while trying to fetch the routes!

showcase/src/assets/examples/route-search/route-search-with-param.ts
```diff
@@ -0,0 +1,105 @@
+import { Component, Inject, OnInit } from "@angular/core";
+import { STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
+import { StarkMenuConfig } from "../../../../../packages/stark-ui";
```
christophercr: Wrong import

showcase/src/styles/_stark-styles.scss
```diff
@@ -34,9 +34,11 @@ IMPORTANT: Stark styles are provided as SCSS styles so they should be imported i
 @import "~@nationalbankbelgium/stark-ui/src/modules/table/components/table-theme";
 @import "~@nationalbankbelgium/stark-ui/src/modules/table/components/dialogs/multisort.component";
 @import "~@nationalbankbelgium/stark-ui/src/modules/dropdown/components/dropdown-theme";
+@import "~@nationalbankbelgium/stark-ui/src/modules/route-search/components/route-search-theme";
 @import "~@nationalbankbelgium/stark-ui/src/modules/toast-notification/components/toast-notification.component";
 @import "~@nationalbankbelgium/stark-ui/src/modules/toast-notification/components/toast-notification-theme";
 @import "~@nationalbankbelgium/stark-ui/src/modules/dropdown/components/dropdown.component";
+@import "~@nationalbankbelgium/stark-ui/src/modules/route-search/components/route-search.component";
```
christophercr: Can you put the imports for the same component together?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		{
+			return "";
+		}
+	}
+
+	/**
+	 * Retrieve the list of routes from a StarkMenuConfig object which the user have passed to the component
+	 */
+	//tslint:disable:cognitive-complexity
+	public initRoutePathFromConfig(): RouteEntry[]{
```
christophercr: Shouldn't this function be recursive? Because the StarkMenuConfig is recursive as well...

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		const optionsToDisplay: RouteEntry[] = [];
+		if (this.optionsList !== undefined && this.optionsList.menuSections !== undefined) {
+			this.optionsList.menuSections.forEach((section: StarkMenuSection) => {
+				section.menuGroups.forEach((group: StarkMenuGroup) => {
```
christophercr: We should use `for ... of` loops instead of `Array.forEach()`.

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+					if (group.isVisible && group.isEnabled && group.targetState) {
+						optionsToDisplay.push({
+							label: group.label,
+							targetState: group.targetState,
+							targetStateParams: group.targetStateParams
+						});
+					}
+					if (group.entries) {
+						group.entries.forEach((entry: StarkMenuGroup) => {
```
christophercr: We should use `for ... of` loops instead of `Array.forEach()`.

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+			}
+		}
+		return optionsToDisplay;
+	}
+
+	/**
+	 * Sorts the list of path alphabetically 
+	 */
+	public sortPathList():void{
+		this.optionsToDisplay.sort((firstEntry: RouteEntry,secondEntry: RouteEntry) => {
```
christophercr: Couldn't we use [Lodash sortBy()](https://lodash.com/docs#sortBy) for this?

packages/stark-ui/src/modules/route-search/entities/route-entry.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Rout
```
christophercr: Can you put something more descriptive here?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+	public inputValue: string ="";
+	
+	/**
+	 * if the input field must be hidden or not
+	 */
+	public hide: boolean;
+
+	public constructor(@Inject(STARK_ROUTING_SERVICE) public starkRoutingService: StarkRoutingService,
+					   @Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService,
+					   @Inject(UIRouter) public routingService: UIRouter) {
```
christophercr: You don't need the `UIRouter`to get all the information about the states. You can get the same information from the `StarkRoutingService`.

packages/stark-ui/src/modules/route-search/components/route-search.component.html
```diff
@@ -0,0 +1,57 @@
...
+		</button>
+		<div *ngIf="!hide">
+			<mat-form-field floatLabel="never" class="path-input-field">
+				<input class="path-input"
+					   matInput
+					   [matTooltip]="'STARK.ROUTE_SEARCH.PLACEHOLDER'| translate"
+					   [matAutocomplete]="auto"
+					   [placeholder]="'STARK.ROUTE_SEARCH.PLACEHOLDER'| translate"
+					   [(ngModel)]="inputValue"
+					   [formControl]="pathCtrl">
```
christophercr: We should use either `ngModel` or `formControl` but not both... otherwise we might have some weird side effects and potential errors...

packages/stark-ui/src/modules/route-search/components/route-search.component.html
```diff
@@ -0,0 +1,57 @@
...
+	<form class="route-search">
+		<div *ngIf="!hide">
+			<mat-form-field floatLabel="never" class="path-input-field">
+				<input class="path-input"
+					   matInput
+					   [matTooltip]="'STARK.ROUTE_SEARCH.PLACEHOLDER'| translate"
+					   [matAutocomplete]="auto"
+					   [placeholder]="'STARK.ROUTE_SEARCH.PLACEHOLDER'| translate"
+					   [(ngModel)]="inputValue"
+					   [formControl]="pathCtrl">
```
christophercr: We should use either `ngModel` or `formControl` but not both... otherwise we might have some weird side effects and potential errors...

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,199 @@
...
+		this.pathCtrl = new FormControl();
+	}
+	
+	/**
+	 * Filters the options according to what has been entered by the user in the input field
+	 */
+	public filterOptions(value: string): RouteEntry[] {
```
christophercr: Can you rename this method to `filterRouteEntries`... to make it consistent with what the component does

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
christophercr: Can you rename this property to `filteredRouteEntries` to make it clear the intention of this variable?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,187 @@
...
+				}
+			}
+		}
+		return routesToDisplay;
+	}
+
+	/**
+	 * Retrieve the list of routes from UIRouter if the user haven't passed a list to the component
+	 */
+	public initRoutePathFromRouter(): RouteEntry[]{
```
christophercr: Can you rename this to `constructRouteEntriesFromRouterStates` so it is more descriptive?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,187 @@
...
+ * Name of the component
+ *
+ */
+const componentName: string = "route-search";
+
+/**
+ * Component to search a particular route with autocompletion that will redirect to the route chosen by the user
+ */
+@Component({
+	selector: "route-search",
```
christophercr: ⚠️ don't forget the `stark` prefix

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,187 @@
...
+import {map, startWith} from "rxjs/operators";
+import { STARK_ROUTING_SERVICE, StarkRoutingService, STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
+import { StarkMenuConfig } from "../../app-menu/components";
+const _sortBy : Function = require("lodash/sortBy");
+import { StateDeclaration } from "@uirouter/core";
+/**
+ * Name of the component
+ *
+ */
+const componentName: string = "route-search";
```
christophercr: ⚠️ don't forget the `stark` prefix

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * Retrieve the list of routes from UIRouter if the user haven't passed a list to the component
+	 */
+	public initRoutePathFromRouter(): RouteEntry[]{
+		const routesToDisplay: RouteEntry[] = [];
+		if (this.menuConfig === undefined){
+			for (const state of this.starkRoutingService.getStatesConfig()) {
+				console.log(state);
+				if (state.name !== undefined && state.url !== undefined && state.url.length > 1) {
+					if (state.url.startsWith("^/")) {
```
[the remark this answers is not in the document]
Mallikki: this block was made to display the route a bit nicer in the component

Mallikki:
So, I've made my last commit, the changes pass now, but I still don't know how to make that function recursive. I asked the help of @tenretC on this, because he has done that before, but we don't get how it can be recursive has the attributes changes from one group to another: in the sections, we check for groups and in groups, we check for entries. We don't check the same thing, so how could we get out of the function? 

The rest is all fine, the tests too

christophercr:
Closed in favor of #1082

PR #980, opened 2018-12-21
feat(stark-ui): make dropdown component working with reactive form + add stark form util

SuperITMan:
ISSUES CLOSED: #982 

## PR Checklist

## PR Type

```
[X] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[X] Documentation content changes
[ ] Other... Please describe:
```

The dropdown component has no support for Reactive form.

Issue Number: #982 

## What is the new behavior?

Dropdown component is based only on Reactive Form but still accepts `value` + `selectionChanged` tobe able to use it. The developer can define himself the formControl object by passing via the input `dropdownFormControl`.

Form utils have been added to Stark.

```
[X] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
+++ b/packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
@@ -1,24 +1,39 @@
 import {
 	Component,
+	ElementRef,
 	EventEmitter,
 	Inject,
 	Input,
 	OnChanges,
+	OnDestroy,
 	OnInit,
 	Output,
+	Renderer2,
 	SimpleChanges,
-	ViewEncapsulation,
-	ElementRef,
-	Renderer2
+	ViewEncapsulation
 } from "@angular/core";
 import { STARK_LOGGING_SERVICE, StarkLoggingService } from "@nationalbankbelgium/stark-core";
 import { AbstractStarkUiComponent } from "../../../common/classes/abstract-component";
-import { StarkComponentUtil } from "../../../util/component";
+import { StarkComponentUtil, StarkFormUtil } from "../../../util";
+import { FormControl, Validators } from "@angular/forms";
+import { distinctUntilChanged, filter } from "rxjs/operators";
+import { Subscription } from "rxjs";
+
+/**
+ * @ignore
+ */
+const _isEqual: Function = require("lodash/isEqual");
 
 /**
  * Name of the component
  */
 const componentName: string = "stark-dropdown";
+
+/**
+ * Warning message for reactive form issue
+ */
+const reactiveWarningMessage: string =
+	componentName + ': When dropdownFormControl is set, "value", "selectionChanged" and "isDisabled" should not be defined.';
 
 /**
  * Component to display dropdown list based on the options passed as parameters. The dropdown component is based
@@ -33,12 +48,18 @@ const componentName: string = "stark-dropdown";
 		class: componentName
 	}
 })
-export class StarkDropdownComponent extends AbstractStarkUiComponent implements OnInit, OnChanges, OnInit {
+export class StarkDropdownComponent extends AbstractStarkUiComponent implements OnInit, OnChanges, OnInit, OnDestroy {
 	/**
 	 * If the dropdown will contain a default blank (optional)
 	 */
 	@Input()
 	public defaultBlank?: boolean;
+
+	/**
+	 * FormControl object to be used in the dropdown.
+	 */
+	@Input()
+	public dropdownFormControl?: FormControl;
 
 	/**
 	 * HTML "id" attribute of the element.
@@ -64,13 +85,6 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements 
 	 */
 	@Input()
 	public isDisabled?: boolean = false;
-
-	/**
-	 * Text to be displayed as the dropdown's label (dynamically translated via the $translate service if
-	 * the provided text is defined in the translation keys).
-	 */
-	@Input()
-	public label?: string;
 
 	/**
 	 * Allows multiple option selection. Setting the attribute to "true" or empty
@@ -118,13 +132,13 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements 
 	 * Source object to be bound to the dropdown ngModel.
 	 */
 	@Input()
-	public value: any | any[];
+	public value?: any | any[];
 
 	/**
 	 * This will emit the newly selected value.
 	 */
 	@Output()
-	public dropdownSelectionChanged: EventEmitter<any> = new EventEmitter<any>();
+	public selectionChanged?: EventEmitter<any> = new EventEmitter<any>();
 
 	/**
 	 * Whether the multiple row selection is enabled.
@@ -136,6 +150,17 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements 
 	 * @internal
 	 */
 	public optionsAreSimpleTypes: boolean;
+
+	/**
+	 * @ignore
+	 */
+	public formControl: FormControl;
+
+	/**
+	 * @ignore
+	 * @internal
+	 */
+	private formControlSubscription: Subscription;
 
 	/**
 	 * Class constructor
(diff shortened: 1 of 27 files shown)


packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -1,25 +1,35 @@
...
 /**
  * Name of the component
  */
 const componentName: string = "stark-dropdown";
 
+/**
+ * Warning message for reactive form issue
+ */
+const reactiveWarningMessage: string =
+	componentName + ': When dropdownFormControl is set, "value", "selectionChanged" and "isDisabled"should be undefined.';
```
christophercr: Missing whitespace... and small change: "When dropdownFormControl is set, "value", "selectionChanged" and "isDisabled" should not be defined."

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -172,13 +222,60 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements
 		if (changes["multiSelect"]) {
 			this.isMultiSelectEnabled = StarkComponentUtil.isInputEnabled(this.multiSelect);
 		}
+
+		if (changes["isDisabled"]) {
+			if (typeof this.dropdownFormControl !== "undefined") {
```
christophercr: Maybe you can use the type-guard you use in the `ngOnInit` here instead: `StarkFormUtil.isFormControl(this.dropdownFormControl)` so we really check for a formControl

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -172,13 +222,60 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements
...
+					this.formControl.enable();
+				}
+			}
+		}
+
+		if (changes["required"]) {
+			this.setDefaultBlank();
+			if (
+				!changes["required"].firstChange &&
+				typeof this.dropdownFormControl === "undefined" &&
```
christophercr: Same remark: maybe you can use the type-guard you use in the ngOnInit here instead

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -172,13 +222,60 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements
...
+				} else {
+					this.formControl.clearValidators();
+				}
+
+				this.formControl.updateValueAndValidity();
+			}
+		}
+
+		if (changes["value"]) {
+			if (typeof this.dropdownFormControl !== "undefined") {
```
christophercr: Same remark: maybe you can use the type-guard you use in the ngOnInit here instead

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
christophercr: In fact, why we need to check for the `dropdownFormControl`? At the end we only use the `formControl` right?

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -172,13 +222,60 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements
...
+			this.setDefaultBlank();
+			if (
+				!changes["required"].firstChange &&
+				typeof this.dropdownFormControl === "undefined" &&
+				typeof this.formControl !== "undefined"
+			) {
+				if (this.required) {
+					this.formControl.setValidators([Validators.required]);
+				} else {
+					this.formControl.clearValidators();
```
christophercr: I don't think we should remove all validators... it could be that the developer added more validators to the form control... we should find a way to remove just the "required" validator but still keep the other ones...

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
```diff
@@ -172,13 +222,60 @@ export class StarkDropdownComponent extends AbstractStarkUiComponent implements
...
+		if (changes["value"]) {
+			if (typeof this.dropdownFormControl !== "undefined") {
+				this.logger.warn(reactiveWarningMessage);
+			}
+
+			if (
+				!changes["value"].firstChange &&
+				typeof this.value !== "undefined" &&
+				typeof this.formControl !== "undefined" &&
+				((Array.isArray(this.value) && this.formControl.value + "" !== this.value.toString()) ||
```
christophercr: Just to confim... is this working fine? The `toString()` and the `string + ""` give the same results? does it work in IE and Firefox? Cause I have the feeling that this really depends on the Javascript engine of the browser...

packages/stark-ui/src/util/form/form.util.ts
```diff
@@ -0,0 +1,122 @@
...
+export type StarkFormState = "untouched" | "pristine" | "submitted" | "dirty";
+
+export class StarkFormUtil {
+	/**
+	 * Set all the fields of the form group to the given state(s)
+	 * @param formGroup - Angular form group object
+	 * @param statesToBeSet - States to be set to the different controls in the form
+	 */
+	public static setFormChildControlsState(formGroup: FormGroup, statesToBeSet: StarkFormControlState[]): void {
+		// Verifying it is indeed an Angular FormController
```
christophercr: `FormController` -> `FormGroup`

packages/stark-ui/src/util/form/form.util.ts
```diff
@@ -0,0 +1,122 @@
...
+		if (StarkFormUtil.isFormGroup(formGroup)) {
+			for (const key of Object.keys(formGroup.controls)) {
+				// filtering just the ngModel child objects of the form
```
christophercr: `ngModel child objects` -> `FormControl objects`

packages/stark-ui/src/util/form/form.util.ts
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	/**
+	 * Set the form to the given state(s). Possible states: "untouched", "pristine", "dirty", "submitted"
+	 * @param formGroup - Angular form group object
+	 * @param statesToBeSet - States to be set to the form
+	 */
+	public static setFormGroupState(formGroup: FormGroup, statesToBeSet: StarkFormState[]): void {
+		// Verifying it is indeed an Angular FormController
```
christophercr: `FormController` -> `FormGroup`

packages/stark-ui/src/util/form/form.util.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+			const result: boolean = StarkFormUtil.isFormGroupValid(mockFormGroup);
+
+			expect(result).toBe(false);
+			expect(StarkFormUtil.setFormChildControlsState).toHaveBeenCalledTimes(1);
+			expect(StarkFormUtil.setFormChildControlsState).toHaveBeenCalledWith(mockFormGroup, ["touched"]);
+		});
+	});
+
+	describe("isFormGroup", () => {
+		it("should return TRUE if the form is an instance of an IFormController", () => {
```
christophercr: `IFormController` -> `FormGroup`

packages/stark-ui/src/util/form/form.util.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+			const result: boolean = StarkFormUtil.isFormGroup(mockFormGroup);
+
+			expect(result).toBe(true);
+		});
+
+		it("should return FALSE if the form is NOT an instance of an IFormController", () => {
```
christophercr: `IFormController` -> `FormGroup`

showcase/src/app/demo-ui/pages/dropdown/demo-dropdown-page.component.html
```diff
@@ -1,6 +1,27 @@
...
+			[options]="serviceDropdownOptions"
+			[dropdownFormControl]="serviceFormControl"
+			placeholder="SHOWCASE.DEMO.DROPDOWN.SELECT_SERVICE"
+			optionIdProperty="id"
+			optionLabelProperty="value"
+			required
+		>
+		</stark-dropdown>
+
+		<div *ngIf="selectedServiceFormControl">
```
christophercr: For this example, can you also demo how it can be disabled/enabled via the form control?

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
SuperITMan: good catch, ty

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
SuperITMan: In this case, we have the control on the `formControl`, so we can actually do it. 
In any cases, Angular does not provide a way to remove only one validator. It is all or nothing.
You can see this StackOverFlow post about this (they provide a kind of a solution): https://stackoverflow.com/questions/46488078/angular-4-remove-required-validator-conditionally 

If the developer defines a dropdownFormControl, the logic to add/remove the `required` validator to the `formControl` object is not done. That means we can maintain the logic as it is until we need to play with more than one validator internally.

showcase/src/app/demo-ui/pages/dropdown/demo-dropdown-page.component.html
SuperITMan: Good idea 😊

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
christophercr: Ah true, this logic is only when the developer does not provide a form control... my bad.

About the `clearValidators; indeed you're right, Angular does not provide such feature but for the moment we don't need it :p

packages/stark-ui/src/modules/dropdown/components/dropdown.component.ts
SuperITMan: I solved this issue by using isEqual method from lodash 😊

Issue #1053, opened 2019-01-17
all: set dependencies version to accept next patches or next minor releases ?


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Every day, Greenkeeper creates pull requests to update a dependency from version A to version B. Often, those updates are simple patches. 

## Expected behavior

Could we reduce the number of pull requests in Stark by setting the dependencies with `~` to accept next patch versions or `^` to accept next minor versions without having a PR for this ?

Also, because of the fixed versions in our packages, the projects need to upgrade Starl every time they want to have the latest dependencies versions.

SuperITMan:
The goal is to not be blocking for the final users.

So, for **stark packages**, we should accept `^` but we should pay attention if we can really do it for every dependencies (ie: be careful with `@angular-devkit/build-angular` its own dependencies which can sometimes change).

So, as we talked, we should do the following for our **packages** and for **root**: 
* dependencies: use the `^` for every dep (and adapt it in case a build fails as you mentionned)
* devDependencies: use `^` for every dep (and adapt it in case a build fails as you mentionned)
* peerDependencies: use `^7.0.0` instead of `7.x` and adapt the minimum required version based on that. If we support Angular 7 and 8 (because we don't use yet features from Angular 8), put `^7.0.0 || ^8.0.0`. Again, we don't want to block or spam the developer with wrong warning messages during the install.
Also, in this case for Typescript, we should have the peerDependency: `"typescript": "~3.1.1"` instead of `"typescript": ">=3.1.1 && <3.2"`
Of course, we should also adapt the peerDependencies version in case of a build fails.

For **starter** and **showcase**, I agree with what you propose 😊

PR #1079, opened 2019-01-24
feat(stark-all): prefix all (dev)dependencies with '^' in all Stark's packages + in starter & showcase

SuperITMan:
The goal is to reduce the number of GreenKeeper's pull requests on the repo

ISSUES CLOSED: #1053

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Dependencies are fixed in Stark's packages and can annoy the final developer with useless warnings.
Also, that forces us to make a new release of Stark everytime we want to let the developers having the latest versions of our dependencies. 

Issue Number: #1053 

## What is the new behavior?
We are not blocking the developer with fixed dependencies.
We are depending on a specific version (and the next minor ones) for each dependency.

Now, every day when Travis runs a build, that will check if the dependencies are still OK and we'll adapt the versions of our dependencies if needed.

```
[ ] Yes
[X] No
```

## Other information


christophercr:
@cnomes indeed we will set all angular deps to "^x.0.0" 😉 

About the `peerDependencies`, indeed, when a dependency is explicitly used in your code you should put it in `dependencies` but our case is special because we depend on Angular (a framework) and Typescript (a language) so those dependencies are always needed in an application even if you don't use Stark.

Moreover, although Stark needs Angular and Typescript we cannot add them to our `dependencies` because it is the developer that decides which version of Angular to install.

Therefore we only set them as `peerDependencies` with the minimum version we know that Stark is compatible with. So in case the user installs a version out of the range of our `peerDependencies` there will be a warning when he runs `npm install`.

PR #1082, opened 2019-01-25
feat(stark-ui): implement Route Search component

christophercr:
ISSUES CLOSED: #207

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #207 

## What is the new behavior?
A route search component has been implemented to allow the user to search for a specific page and navigate to it directly.

```
[ ] Yes
[X] No
```


SuperITMan:
I tested it and I have some small remarks when I tested it about the current implementation. 

The "g" is know displayed correclty in the input:
image

When we enter a route name, it is not auto-selected and so, we have to use "bottom" key or move the mouse on the element and then we can type on "Enter" to access the route. 
image

When we click on the button to show the route-search input to type the route name, we have to click on the input after it has been opened to be able to type the route name. Could we have an auto-blur for this ? 😊 

In the showcase, when we look at the code, we set the input `direction="right"`, but, as you can see above, the input is shown on the left. Is the **direction** working ?

Finally, still for our usage of the route-search in the showcase, could we pass our menuConfig for having all the available routes in the component ? 😊

christophercr:
@SuperITMan I've updated the PR with the fixes for the remarks ;)

Could you please have a look?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,287 @@
...
+
+	/**
+	 * The list of {@link RouteEntry}'s objects filtered by the auto-complete field
+	 */
+	public filteredRouteEntries: Observable<RouteEntry[]>;
+
+	/**
+	 * The list of routes to be displayed
+	 */
+	public routesToDisplay: RouteEntry[] = [];
```
SuperITMan: Could we create an `Input()` for this to let the developer passing its own `routesToDisplay` if he does not use stark-app-menu and so does not want to build a `StarkMenuConfig` object instead of a simple array of objects with only for 3 properties ? 😊

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,287 @@
...
+			map((option: string) => {
+				return option ? this.filterRouteEntries(option) : [...this.routesToDisplay];
+			})
+		);
+
+		this.hide = true;
+
+		this.logger.debug(componentName + ": component initialized");
+
+		if (this.routesToDisplay !== undefined) {
```
SuperITMan: According to what I see here, I guess you already thought to set `routesToDisplay` as an Input binding 😊

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,287 @@
...
+			if (this.menuConfig === undefined) {
```
SuperITMan: Would not it better to check `typeof this.menuConfig === "undefined"` ?

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
```diff
@@ -0,0 +1,287 @@
...
+				routesToDisplay.push({ label: translationKey, targetState: ng2State.name });
+			}
+		}
+
+		return routesToDisplay;
+	}
+
+	/**
+	 * Translate the label of all elements in the {@link RouteEntry} object.
+	 * This is useful as the language can change during runtime.
```
SuperITMan: If we simply put the translation in the HTML like this:

```html
<mat-option
	(onSelectionChange)="redirect(option)"
	*ngFor="let option of (filteredRouteEntries | async); trackBy: trackPath"
	[value]="option.label  | translate"
>
	<p>{{ option.label | translate }}</p>
</mat-option>
```

That means it won't translated if the language changes during runtime ?

christophercr: In fact, the translation is not the problem. The problem is the sorting because they options should be sorted using the translated values so it cannot be done only in the template cause there we only have the translation keys ;)

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
christophercr: As discussed, we'll leave that as a "nice to have" for later ;)

packages/stark-ui/src/modules/route-search/components/route-search.component.ts
christophercr: As discussed, we'll leave that as a "nice to have" for later ;)

christophercr:
Changes applied. Can you review it again please?

Issue #1130, opened 2019-02-12
core: http - HttpClient does not encode '+' char properly (known bug)


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

The `+` char is not correctly encoded in the query string of the URL of the HTTP call.

This is already a known bug in the Angular HttpClient: https://github.com/angular/angular/issues/18261 and https://github.com/angular/angular/issues/11058 which remain open for the last couple of years.

## Expected behavior

The URL of every HTTP call should be correctly encoded.

## Minimal reproduction of the problem with instructions

- Trigger an HTTP request with a query parameter with a date string. For example: `dateFrom:2018-01-01T00:00:00+01:00`
- The HTTP call should have the parameter encoded as: `dateTo=2018-01-01T00:00:00%2B01:00` but it is not encoded at all: `dateFrom=2018-01-01T00:00:00+01:00 `. The `+` is not encoded.

## What is the motivation / use case for changing the behavior?

Respect Http standards.

## Environment

<pre><code>
Angular version: 7.2.3
Stark version: 10.0.0-beta.3

</code></pre>

christophercr:
The [solution suggested in the threads](https://github.com/angular/angular/issues/18261#issuecomment-426383787) is to create a [HttpParameterCodec](https://angular.io/api/common/http/HttpParameterCodec).

So we should implement that in stark-core.

PR #1132, opened 2019-02-13
fix(stark-core): use custom encoder for URL query parameters

cnomes:
- extracted `convertMapIntoObject` function to separate file
  - added `convertStarkQueryParamsIntoHttpParams` function to correctly handle params in the http module
  - extended `HttpParameterCodec` according to [this solution](https://github.com/angular/angular/issues/18261#issuecomment-426383787)
  - updated `http.util.spec.ts` according to new implementation
  - test `convertStarkQueryParamsIntoHttpParams`

ISSUES CLOSED: #1130

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1130 

## What is the new behavior?
Encodes special characters correctly

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-core/src/modules/http/services/http.service.spec.ts
+++ b/packages/stark-core/src/modules/http/services/http.service.spec.ts
@@ -5,9 +5,12 @@ import SpyObj = jasmine.SpyObj;
 import { autoserialize, autoserializeAs, inheritSerialization, Serialize } from "cerialize";
 import { Observable, of, throwError } from "rxjs";
 import { catchError } from "rxjs/operators";
-import { HttpClient, HttpErrorResponse, HttpHeaders, HttpResponse } from "@angular/common/http";
-
+import { HttpClient, HttpErrorResponse, HttpHeaders, HttpParams, HttpResponse } from "@angular/common/http";
+import { convertMapIntoObject } from "../../../util/util-helpers";
+import { StarkHttpUtil } from "../../../util/http.util";
 import { StarkHttpServiceImpl } from "./http.service";
+import { MockStarkLoggingService } from "../../logging/testing";
+import { MockStarkSessionService } from "../../session/testing";
 import {
 	StarkBackend,
 	StarkBackendImpl,
@@ -32,8 +35,6 @@ import {
 
 import { StarkHttpHeaders, StarkSortOrder } from "../constants";
 import { StarkHttpStatusCodes } from "../enumerators";
-import { MockStarkLoggingService } from "../../logging/testing";
-import { MockStarkSessionService } from "../../session/testing";
 import { StarkHttpSerializer, StarkHttpSerializerImpl } from "../serializer";
 
 /* tslint:disable:no-big-function no-duplicate-string max-union-size */
@@ -149,9 +150,7 @@ describe("Service: StarkHttpService", () => {
 	};
 
 	interface HttpRequestOptions {
-		params: {
-			[param: string]: string | string[];
-		};
+		params: HttpParams;
 		headers: {
 			[header: string]: string | string[];
 		};
@@ -471,7 +470,7 @@ describe("Service: StarkHttpService", () => {
 							httpMockMethod,
 							mockResourceFullUrl,
 							{
-								params: convertMapIntoObject(request.queryParameters),
+								params: StarkHttpUtil.convertStarkQueryParamsIntoHttpParams(request.queryParameters),
 								headers: convertMapIntoObject(headersMap),
 								observe: "response",
 								responseType: "json"
@@ -565,7 +564,7 @@ describe("Service: StarkHttpService", () => {
 							httpMockMethod,
 							mockResourceFullUrl,
 							{
-								params: convertMapIntoObject(request.queryParameters),
+								params: StarkHttpUtil.convertStarkQueryParamsIntoHttpParams(request.queryParameters),
 								headers: convertMapIntoObject(headersMap),
 								observe: "response",
 								responseType: "json"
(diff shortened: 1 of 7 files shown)


packages/stark-core/src/modules/http/helpers/http-helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+ * A custom implementation of HttpParameterCodec to correctly encode query parameters.
+ * @link https://github.com/NationalBankBelgium/stark/issues/1130
+ */
+export const CUSTOM_ENCODER: HttpParameterCodec = new CustomEncoder();
+
+/**
+ * Converts the Map<string, StarkQueryParam> required by the service into a HttpParams object required by Angular
+ * @param starkQueryParam - params to convert
+ */
+export function convertStarkQueryParamsIntoHttpParams(starkQueryParam: Map<string, StarkQueryParam>): HttpParams {
```
christophercr: I agree this function is something that can be re-used, however I think we should put it as a static method in a Util class (like we the ones we have under `packages/stark-core/src/util/`). This way this can be also used by developers when dealing with `StarkQuerrParams`

packages/stark-core/src/modules/http/helpers/http-helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+	return _reduce(
+		convertMapIntoObject(starkQueryParam), // convert to object
+		(httpParams: HttpParams, starkParams: StarkQueryParam, key: string) =>
+			starkParams
+				? Array.isArray(starkParams)
+					? _reduce(starkParams, (acc: HttpParams, entry: string) => acc.append(key, entry), httpParams)
+					: httpParams.set(key, starkParams)
+				: httpParams.set(key, ""),
```
christophercr: Why is this key set to ""?

packages/stark-core/src/modules/http/helpers/http-helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Converts the Map<string, StarkQueryParam> required by the service into a HttpParams object required by Angular
+ * @param starkQueryParam - params to convert
+ */
+export function convertStarkQueryParamsIntoHttpParams(starkQueryParam: Map<string, StarkQueryParam>): HttpParams {
+	return _reduce(
+		convertMapIntoObject(starkQueryParam), // convert to object
+		(httpParams: HttpParams, starkParams: StarkQueryParam, key: string) =>
+			starkParams
```
christophercr: Can you use `if` blocks here since we have nested blocks and it is more diffictul to read the code?

packages/stark-core/src/modules/http/helpers/http-helpers.ts
```diff
@@ -0,0 +1,45 @@
...
+				: httpParams.set(key, ""),
+		new HttpParams({ encoder: CUSTOM_ENCODER })
+	);
+}
+
+/**
+ * Converts a map (used for params and headers) into a plain object
+ * @param map - the Map to convert
+ */
+export function convertMapIntoObject(map: Map<string, any>): { [param: string]: any } {
```
christophercr: Indeed, this function is better here so we can re-use it ;)

packages/stark-core/src/modules/http/services/http.service.spec.ts
```diff
@@ -149,9 +150,11 @@ describe("Service: StarkHttpService", () => {
...
 
 	interface HttpRequestOptions {
-		params: {
-			[param: string]: string | string[];
-		};
+		params:
+			| {
+					[param: string]: string | string[];
+			  }
+			| HttpParams;
```
christophercr: Shouldn't the type of this property be simply `params: HttpParams` ?

packages/stark-core/src/modules/http/helpers/http-helpers.ts
cnomes: So should i move it to there or create a util class specific to http?

packages/stark-core/src/modules/http/helpers/http-helpers.ts
cnomes: set is defined as `set(param: string, value: string): HttpParams;` so when it's `undefined` I set it to an empty string

packages/stark-core/src/modules/http/helpers/http-helpers.ts
christophercr: Indeed, we should create a Util class specific for Http ;)

packages/stark-core/src/modules/http/helpers/http-helpers.ts
cnomes: I like actually really like this pattern. But maybe it's clearer like this?
```
typeof value === "undefined" // set key to empty string when not defined
	? httpParams.set(key, "")
	: Array.isArray(value) // append each string to the key when set to an array
	? _reduce(value, (acc: HttpParams, entry: string) => acc.append(key, entry), httpParams)
	: httpParams.set(key, value),
```

packages/stark-core/src/modules/http/services/http.service.spec.ts
cnomes: The complete angular options are:
```
options?: {
    headers?: HttpHeaders | {[header: string]: string | string[]},
    observe?: 'body',
    params?: HttpParams|{[param: string]: string | string[]},
    reportProgress?: boolean,
    responseType?: 'json',
    withCredentials?: boolean,
  }
```

But it seems less generic and more specific to our implementation. So I think you are right. Maybe I should also rename it to something like `StarkHttpRequestOptions`? 🤔

christophercr: Now that I look at it in more detail, the way you have it now is correct... because it matches the typings from Angular HttpClient... so never mind... don't care about this comment ;)

cnomes:
@christophercr I updated the PR.

Issue #1087, opened 2019-01-29
ui: table - When data is NOT set at the initialization, the table component is NOT built correctly


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

If no data is set at the initialization of `stark-table`, the component is broken and changes on `data` are not applied.
```html
<stark-table [data]="undefined"></stark-table>
```

## Expected behavior

The `stark-table` component should work even if there is no data at the initialization time.

PR #1168, opened 2019-03-04
fix(stark-ui): fix table error when `data` not initialized

cnomes:
- added test
    - refactored test typing

ISSUES CLOSED: #1087

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1087 

## What is the new behavior?
The `stark-table` does not fail on undefined data anymore

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/table/components/table.component.spec.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.spec.ts
@@ -49,8 +49,6 @@ class TestHostComponent {
 	public tableComponent: StarkTableComponent;
 
 	public columnProperties: StarkTableColumnProperties[];
-	// FIXME Currently, if no data is set on the init of the stark-table component, it breaks the component and its tests.
-	// More info on: https://github.com/NationalBankBelgium/stark/issues/1087
 	public dummyData: object[] = [];
 	public fixedHeader?: string;
 	public rowsSelectable?: boolean;
@@ -1103,6 +1101,34 @@ describe("TableComponent", () => {
 			hostFixture.detectChanges();
 
 			expect(component.selectChanged.emit).toHaveBeenCalledWith([dummyData[0]]);
+		});
+	});
+
+	describe("async", () => {
+		const dummyData: object[] = [
+			{ id: 1, description: "dummy 1" },
+			{ id: 2, description: "dummy 2" },
+			{ id: 3, description: "dummy 3" }
+		];
+
+		beforeEach(() => {
+			hostComponent.columnProperties = [{ name: "id" }, { name: "description" }];
+			hostComponent.dummyData = <any>undefined; // data starts uninitialized
+
+			hostFixture.detectChanges(); // trigger data binding
+			component.ngAfterViewInit();
+		});
+
+		it("should update rows after async data fetch", () => {
+			const rowsBeforeData: NodeListOf<HTMLTableRowElement> = hostFixture.nativeElement.querySelectorAll("table tbody tr");
+			expect(rowsBeforeData.length).toBe(0);
+
+			// "async fetch of data resolves"
+			hostComponent.dummyData = dummyData;
+			hostFixture.detectChanges();
+
+			const rowsAfterData: NodeListOf<HTMLTableRowElement> = hostFixture.nativeElement.querySelectorAll("table tbody tr");
+			expect(rowsAfterData.length).toBe(dummyData.length);
 		});
 	});
 });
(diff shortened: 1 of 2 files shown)


packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -1105,4 +1103,36 @@ describe("TableComponent", () => {
 			expect(component.selectChanged.emit).toHaveBeenCalledWith([dummyData[0]]);
 		});
 	});
+
+	describe("async", () => {
+		const dummyData: object[] = [
+			{ id: 1, description: "dummy 1" },
+			{ id: 2, description: "dummy 2" },
+			{
+				id: 3,
```
christophercr: Can you format this block the same as the 2 lines above?

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -1105,4 +1103,36 @@ describe("TableComponent", () => {
...
+				description: "dummy 3"
+			}
+		];
+		console.log(dummyData);
```
christophercr: Can you remove this `console.log`?

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -1105,4 +1103,36 @@ describe("TableComponent", () => {
...
+		it("should update rows after async data fetch", () => {
+			const rowsBeforeData: NodeListOf<HTMLTableRowElement> = hostFixture.nativeElement.querySelectorAll("table tbody tr");
+			expect(rowsBeforeData.length).toBe(0);
+
+			// "async fetch of data resolves"
+			hostComponent.dummyData = dummyData;
+			hostFixture.detectChanges();
+
+			const rowsAfterData: NodeListOf<HTMLTableRowElement> = hostFixture.nativeElement.querySelectorAll("table tbody tr");
+			expect(rowsAfterData.length).toBe(3);
```
christophercr: Maybe it is better to put `dummyData.length` instead of `3` to make it more "dynamic" so this doesn't need to change if we change the dummy data at some point ;)

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -1105,4 +1103,36 @@ describe("TableComponent", () => {
...
+			{
+				id: 3,
+				description: "dummy 3"
+			}
+		];
+		console.log(dummyData);
+
+		beforeEach(() => {
+			hostComponent.columnProperties = [{ name: "id" }, { name: "description" }];
+			hostComponent.dummyData = undefined; // data starts uninitialized
```
christophercr: Maybe it is less code to write if the dummyData is defined as it was before: `public dummyData: object[] = [];` so you don't need to cast to  `<object[]>` every time. You would just need to cast it in this line to `<any>undefined`. What do you think?

cnomes: I updated the PR. 😄

christophercr:
Strangely enough, the build on Travis for Node 8 fails. The one for Node 10 passes.

The error thrown is in the unit tests:

```
 TableComponent
    on initialization
            should set internal component properties[39m
              HeadlessChrome 72.0.3617 (Linux 0.0.0)
                 Uncaught TypeError: Cannot read property 'filter' of undefined thrown
```

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -524,7 +524,7 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	 */
 	// tslint:disable-next-line:cognitive-complexity
 	private initializeDataSource(): void {
-		this.dataSource = new MatTableDataSource(this.data);
+		this.dataSource = new MatTableDataSource(this.data || []);
```
christophercr: I'm wondering... isn't it better to set `this.data` to `[]` inside the `ngOnChanges` only instead of repeating `this.data || []` wherever it is used? Then we will only have one single point to look at in case of errors with the `data` input.

cnomes: Indeed, this brings the fix back to 1 line. 🙈

Issue #1144, opened 2019-02-20
build: typescript - loosen TS configuration to allow type Inference


cnomes:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Currently when coding typing has to be done very explicitly, this could be reduced by allowing (forcing) type inference through linting.
links:
TypeScript [Type Inference](https://www.typescriptlang.org/docs/handbook/type-inference.html)
TSLint: [no-inferrable-types](https://palantir.github.io/tslint/rules/no-inferrable-types/)

## What is the motivation / use case for changing the behavior?

IMHO this would lead to cleaner code.

christophercr:
As discussed we can take the chance to also enable the flag [strictPropertyInitialization](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#strict-class-initialization) in our TS config. We disabled it a while ago because it was not supported by our version of IntelliJ back then.

cnomes:
@christophercr @SuperITMan with the `strictPropertyInitialization` I have to do a lot of refactoring. But I have code-style question before I start. Do we prefer initializing class properties at declaration or in the constructor.
I prefer at declaration, because then there is a lot less jumping around in the code to find the (initial) value of a property. But I wanted to discuss this before I start working on it. 😄

christophercr:
As discussed, we should initialize the properties at the declaration otherwise we will force the developer to implement a constructor just to set the values of the properties even if the class doesn't really need a constructor

PR #1218, opened 2019-03-25
Feature: strictPropertyInitialization / update tslint

cnomes:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[x] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
refactor(stark-rbac): fix stark-rbac package after Strict Property Initialization

- fix tests
- add lint-ts:fix script
  - ran lint-ts:fix
  - fix tslint warnings

CLOSES ISSUES: #1144

refactor(stark-demo): fix showcase and starter after Strict Property Initialization

- other refactors
- fix tests
- update examples
- add lint-ts:fix script
  - ran lint-ts:fix
  - fix tslint warnings

CLOSES ISSUES: #1144

refactor(stark-ui): fix stark-ui package after Strict Property Initialization

- other refactors
- fix tests
- add lint-ts:fix script
  - ran lint-ts:fix
  - fix tslint warnings

CLOSES ISSUES: #1144

refactor(stark-core): fix stark-core package after Strict Property Initialization

- added `!` to all uninitialized properties on serializable classes
- other refactors
- fix tests
- add lint-ts:fix script
  - ran lint-ts:fix
  - fix tslint warnings

CLOSES ISSUES: #1144

feat(build): update TypeScript configuration and TSLint configuration

- enable `strictPropertyInitialization` in `tsconfig.json`
- enable `no-inferrable-types` in `tslint.json`
- require `typedef` for `arrow-call-signature` in `tslint.json`
- no longer require `typedef` in `tslint.json` for:
  - `variable-declaration`
  - `member-variable-declaration`
  - `object-destructuring`
  - `array-destructuring`

ISSUES CLOSED: #1144

# BREAKING CHANGES:

Projects extending the `@nationalbankbelgium/stark-build/tsconfig.json` will possibly break.
Projects extending the `@nationalbankbelgium/stark-build/config/tslint.json` will most likely receive more tslint warnings.

**build breaking**

Preferably this can be resolved by **refactoring the code** for improved quality, but if desired the effects of this update can be reversed.

To fix your TypeScript build add the following line to your `tsconfig.json` under `"compilerOptions"`.

```JSON
  "strictPropertyInitialization": false
```

**tslint-warnings**

To resolve the warnings received by TSLint when running `npm run lint` you can first try **adding the script** below to your `package.json` and running `npm run lint-ts:fix`.

```JSON
"lint-ts:fix": "node --max_old_space_size=4096 ./node_modules/tslint/bin/tslint --config ./tslint.json --project ./tsconfig.json --format codeFrame --fix",
```

**This should fix most if not all of the new warnings you are getting.**

After running the scripts any additional warnings caused by this update can be solved by **adding types**.

_Alternatively you can revert the new TSLint rules by adding the following lines to your `tslint.json` under `"rules"`._

```
"no-inferrable-types": false,
"typedef": [
  true,
  "call-signature",
  // "arrow-call-signature",
  "parameter",
  "arrow-parameter",
  "property-declaration",
  "variable-declaration",
  "member-variable-declaration",
  "object-destructuring",
  "array-destructuring"
]
```


cnomes:
I'm not done yet with this commit. But because it is a lot to review I wanted to make a preliminary PR. 😄

packages/stark-ui/src/modules/pagination/components/pagination.component.ts
```diff
@@ -70,7 +70,7 @@ export class StarkPaginationComponent extends MatPaginator implements OnInit, On
 	 * StarkPaginationConfig object containing main information for the pagination.
 	 */
 	@Input()
-	public paginationConfig: StarkPaginationConfig;
+	public paginationConfig: StarkPaginationConfig = { totalItems: 0 };
```
christophercr: This variable can just be initialized as `{}` cause we already set the `totalItems = 0` in the `normalizePaginationConfig()` method ;)

So we just set the value in one place

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
```diff
@@ -481,7 +481,6 @@ describe("PrettyPrintComponent", () => {
 
 			it(shouldHaveInputs, () => {
 				expect(component.data).toBe(rawTypescriptData);
-				expect(component.format).toBe("unknownformat");
```
christophercr: Why was this check removed?

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
```diff
@@ -49,13 +57,13 @@ const directiveName: string = "[starkProgressIndicator]";
 })
 export class StarkProgressIndicatorDirective implements OnInit, OnDestroy {
 	@Input()
-	public starkProgressIndicator: StarkProgressIndicatorConfig;
+	public starkProgressIndicator: StarkProgressIndicatorConfig = DEFAULT_PROGRESS_INDICATOR_CONFIG;
```
christophercr: I don't think defining a default value here is a good solution because it has a side effect: an instance is registered via the `StarkProgressIndicatorService`. This means that for every instance of this directive a new config with topic "undefined" will be registered :(

packages/stark-ui/src/modules/session-ui/components/session-card/session-card.component.ts
```diff
@@ -21,5 +21,5 @@ export class StarkSessionCardComponent {
 	 * The title shown in the header of the card
 	 */
 	@Input()
-	public cardTitle: string;
+	public cardTitle?: string;
```
christophercr: Why don't we initialize it as empty string "" as the other cases similar to this one?

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -54,21 +54,21 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
 	 * Configuration object for the slider instance to be created.
 	 */
 	@Input()
-	public sliderConfig: StarkSliderConfig;
+	public sliderConfig: StarkSliderConfig = { range: { max: [], min: [] } };
```
christophercr: I'm not sure if the slider will work correctly if we define this default range...

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -54,21 +54,21 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
...
+	public sliderId = "undefined";
 
 	/**
 	 * Array of numeric values to be set to the slider.
 	 * For simple sliders with just one handle, the array should contain only one value.
 	 * For range sliders, the array should contain two or more values.
 	 */
 	@Input()
-	public values: number[];
+	public values: number[] = [];
```
christophercr: I'm not sure if this change will cause a regression because it is compared to the `latestUnencodedValues` in the `ngOnChanges`...

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -169,27 +179,26 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
...
 	@ContentChild(TemplateRef)
-	public columnTemplate: TemplateRef<object>;
+	public columnTemplate?: TemplateRef<object>;
 
 	/**
 	 * @ignore
 	 * Internal formControl to manage the filter value of the column
 	 */
-	public _filterFormCtrl: FormControl;
+	public _filterFormCtrl: FormControl = new FormControl();
```
christophercr: I think here is better to use the `!` modifier because we know we will instantiate a new formControl in the `ngOnInit`. Otherwise we will also instantiate a dummy formControl which will also consume some memory and at the end it will never be used.

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -130,29 +130,6 @@ describe("TableComponent", () => {
...
-			expect(hostFixture).toBeDefined();
-			expect(component).toBeDefined();
-
-			expect(component.dialogService).not.toBeNull();
-			expect(component.dialogService).toBeDefined();
-			expect(component.logger).not.toBeNull();
-			expect(component.logger).toBeDefined();
-		});
-
-		it("should NOT have any inputs set", () => {
```
christophercr: I think we should keep this test to be sure what the initial value of the properties is.

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -251,30 +251,30 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
 	@ViewChildren(StarkTableColumnComponent)
-	public viewColumns: QueryList<StarkTableColumnComponent>;
+	public viewColumns!: QueryList<StarkTableColumnComponent>;
 
 	/**
 	 * Columns added by the user via transclusion inside an <div> element with class "stark-table-columns"
 	 */
 	@ContentChildren(StarkTableColumnComponent)
-	public contentColumns: QueryList<StarkTableColumnComponent>;
+	public contentColumns: QueryList<StarkTableColumnComponent> = new QueryList();
```
christophercr: I think we should simply use the `!` modifier here to avoid creating a dummy `QueryList` which might consume some memory which at the end will not be used

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -284,34 +284,34 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
-	public dataSource: MatTableDataSource<object>;
+	public dataSource!: MatTableDataSource<object>;
 
 	/**
 	 * Array of columns (column id's) to be displayed in the table.
 	 */
 	public displayedColumns: string[] = [];
 
-	public _globalFilterFormCtrl: FormControl;
+	public _globalFilterFormCtrl: FormControl = new FormControl();
```
christophercr: Same remark: we should use the `!` modifier to avoid creating a dummy formControl which will never be used

packages/stark-core/src/modules/settings/entities/settings.entity.ts
```diff
@@ -5,5 +5,5 @@ export class StarkSettings {
 	/**
 	 * the user's preferred language
 	 */
-	public preferredLanguage: string;
+	public preferredLanguage?: string;
```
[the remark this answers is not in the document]
cnomes: Not always, at 
`packages/stark-core/src/modules/settings/reducers/settings.reducer.ts(12:0)`
 it will be undefined. 😕 Perhaps I could assign it a default value? (`""`)

packages/stark-core/src/modules/xsrf/services/xsrf.service.spec.ts
```diff
@@ -288,20 +288,22 @@ describe("Service: StarkXSRFService", () => {
 		it("should log an error when the HTTP call to a backend failed", fakeAsync(() => {
 			const failingBackends: StarkBackend[] = [mockBackend1, mockBackend3];
 
-			(<Spy<HttpClientGet>>httpMock.get).and.callFake((url: string) => {
-				if (failingBackends.map((failingBackend: StarkBackend) => failingBackend.url).indexOf(url) !== -1) {
-					return throwError(new HttpErrorResponse({ error: "ping failed" }));
-				} else {
-					return of(new HttpResponse({ body: "ping OK" }));
+			httpMock.get.and.callFake(
+				(url: string): Observable<any> => {
```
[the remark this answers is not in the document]
cnomes: I can't, but I'm not sure why.

I get
```
Error:(292, 5) TS2345: Argument of type '(url: string) => Observable<HttpResponse<string>>' is not assignable to parameter of type '{ (url: string, options: { headers?: HttpHeaders | { [header: string]: string | string[]; } | undefined; observe?: "body" | undefined; params?: HttpParams | { [param: string]: string | string[]; } | undefined; reportProgress?: boolean | undefined; responseType: "arraybuffer"; withCredentials?: boolean | undefined; }...'.
  Type 'Observable<HttpResponse<string>>' is not assignable to type 'Observable<ArrayBuffer>'.
    Type 'HttpResponse<string>' is not assignable to type 'ArrayBuffer'.
      Property 'byteLength' is missing in type 'HttpResponse<string>'.
```

And then maybe I should include the error response. eg: `Observable<HttpResponse<string> | never>`

packages/stark-ui/src/modules/action-bar/components/action-bar.component.ts
```diff
@@ -47,7 +47,7 @@ export class StarkActionBarComponent extends AbstractStarkUiComponent implements
 	 * Alternative actions
 	 */
 	@Input()
-	public alternativeActions: StarkAction[];
+	public alternativeActions: StarkAction[] = [];
```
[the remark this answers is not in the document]
cnomes: It does have the wrong effect. 😅

packages/stark-ui/src/modules/breadcrumb/components/breadcrumb.component.spec.ts
```diff
@@ -201,12 +199,6 @@ describe("BreadcrumbComponent", () => {
 			expect(component.transitionHookDeregisterFn).toBeDefined();
 			expect(component.transitionHookDeregisterFn).toHaveBeenCalledTimes(1);
 		});
-
-		it("should NOT call the transition hook deregistration function if it is undefined", () => {
-			component.ngOnDestroy();
-
-			expect(component.transitionHookDeregisterFn).toBeUndefined();
```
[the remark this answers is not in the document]
cnomes: I updated the tests

packages/stark-ui/src/modules/table/components/column.component.ts
cnomes: Or I remove it from the onInit an change this to `	public _filterFormCtrl: FormControl = new FormControl(this.filterValue);`?

packages/stark-ui/src/modules/table/components/table.component.spec.ts
cnomes: I came across alot of these tests, but I don't really understand them. They don't test anything functional right?

cnomes:
@christophercr @SuperITMan I read this after I made the PR but there is are some Angular guildelines (opinions) on the`StrictPropertyInitialization` https://github.com/angular/angular/issues/24571#issuecomment-404606595

packages/stark-ui/src/modules/table/components/table.component.ts
cnomes: I resolved this similarly to how we discussed the previous comment. 😄

packages/stark-ui/src/modules/slider/components/slider.component.ts
cnomes: What do you suggest I use here? I'm not really familiar with this component and what might be expected behaviour. 😕

SuperITMan:
I checked `stark-core` commit and it seems fine 😊 
One remark, the new rules about type inference should not be set in stark-build but in every subpackage and in tsconfig of showcase.
Why ? Because this is a big breaking change and I don't want to force the projects to change all their code now... 
We'll be able to handle that later in the new https://github.com/nationalbankbelgium/code-style package 😊

cnomes:
> I checked `stark-core` commit and it seems fine 😊
> One remark, the new rules about type inference should not be set in stark-build but in every subpackage and in tsconfig of showcase.
> Why ? Because this is a big breaking change and I don't want to force the projects to change all their code now...
> We'll be able to handle that later in the new https://github.com/nationalbankbelgium/code-style package 😊

# I added following as a commit msg.

refactor(build): disallow explicit definition of inferrable types (tslint)

  - enable strictPropertyInitialization in tsconfig

ISSUES CLOSED: #1144

BREAKING CHANGES:
- projects extending the `@nationalbankbelgium/stark-build/tsconfig.json` and `@nationalbankbelgium/stark-build/config/tslint.json` will most likely break/receive more tslint warnings

- *build-failure*: Preferably this can be resolved by refactoring your code for improved quality, but if desired these effects can be reversed. To fix your TypeScript build add the following line to your `tsconfig.json`.
```
// ...
"compilerOptions":{
  // ...
  "strictPropertyInitialization": false
}
// ...
```

  - *tslint-warnings*: To resolve the warnings received by tslint when running `npm run lint-ts` you can first try adding the following script to your `package.json` and running `npm run lint-ts:fix`. This should fix most if not all of the new warnings you are getting.
```
"lint-ts:fix": "node --max_old_space_size=4096 ./node_modules/tslint/bin/tslint --config ./tslint.json --project ./tsconfig.json --format codeFrame --fix",
```

  - Alternativly you can ignore the new rules by adding the following lines to your `tslint.json`.
```
{
  // ....
  "rules": {
    // ...
    "no-inferrable-types": false,
    "typedef": [
      true,
      "call-signature",
      // "arrow-call-signature",
      "parameter",
      "arrow-parameter",
      "property-declaration",
      "variable-declaration",
      "member-variable-declaration",
      "object-destructuring",
      "array-destructuring"
    ]
  }
```

christophercr:
I've reviewed `stark-core` and it looks good to me 👍

packages/stark-ui/src/modules/breadcrumb/components/breadcrumb.component.ts
```diff
@@ -79,33 +81,31 @@ export class StarkBreadcrumbComponent extends AbstractStarkUiComponent implement
 	 * Component lifecycle hook
 	 */
 	public ngOnDestroy(): void {
-		if (this.transitionHookDeregisterFn) {
-			this.transitionHookDeregisterFn();
-		}
+		this.deregisterTransitionHook();
+		this.deregisterTransitionHook();
```
christophercr: This line is duplicated

packages/stark-ui/src/modules/date-picker/components/date-picker.component.spec.ts
```diff
@@ -114,32 +119,6 @@ describe("DatePickerComponent", () => {
...
-		describe("on initialization", () => {
-			it("should set internal component properties", () => {
-				expect(hostFixture).toBeDefined();
-				expect(component).toBeDefined();
-
-				expect(component.logger).not.toBeNull();
-				expect(component.logger).toBeDefined();
-			});
-
-			it("should NOT have any inputs set", () => {
```
christophercr: I think we should keep this unit test at least to be sure what are the initial values of the different properties...

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -204,35 +208,6 @@ describe("DropdownComponent", () => {
...
 
-		describe("on initialization", () => {
-			it("should set internal component properties", () => {
-				expect(hostFixture).toBeDefined();
-				expect(component).toBeDefined();
-				expect(component.logger).not.toBeNull();
-				expect(component.logger).toBeDefined();
-			});
-
-			it("should NOT have any inputs set", () => {
```
christophercr: Same remark: I think we should keep this unit test at least to be sure the properties are correctly initialized

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -204,35 +208,6 @@ describe("DropdownComponent", () => {
...
-				expect(component.dropdownId).toBe(hostComponent.dropdownId);
-				expect(component.selectionChanged).toBeDefined();
-				expect(component.multiSelect).toBe(!!hostComponent.multiSelect);
-				expect(component.options).toBe(hostComponent.options);
-				expect(component.placeholder).toBe(hostComponent.placeholder);
-				expect(component.disabled).toBeUndefined();
-				expect(component.required).toBe(!!hostComponent.required);
-			});
-
-			it("should render the appropriate content", () => {
```
christophercr: Why was this unit test removed?

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
```diff
@@ -26,12 +26,12 @@ export abstract class AbstractStarkSearchComponent<SearchResultsType, CriteriaTy
 	 * @internal
 	 * @ignore
 	 */
-	private _latestResults: Readonly<SearchResultsType[]>;
+	private _latestResults: Readonly<SearchResultsType[]> = [];
```
christophercr: I don't think that setting an empty array by default has the same effect because then the developer won't have a way to know whether a search has been triggered or not (the results will always be an array). I think we can use the `!` modifier instead because we know that at some point there will be results.

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
```diff
@@ -150,7 +149,10 @@ export abstract class AbstractStarkSearchComponent<SearchResultsType, CriteriaTy
 	 * Invoke the genericSearchService.search() method and emits the results. If no searchCriteria object is passed, then the current
 	 * form's working copy is used.
 	 */
-	public performSearch(searchCriteria: CriteriaType = this.workingCopy): void {
+	public performSearch(searchCriteria: CriteriaType | undefined = this.workingCopy): void {
+		if (!searchCriteria) {
```
christophercr: Why can the `searchCriteria` be undefined? I don't think this if() block is needed either

packages/stark-ui/src/modules/generic-search/components/generic-search/generic-search.component.ts
```diff
@@ -195,31 +195,31 @@ export class StarkGenericSearchComponent extends AbstractStarkUiComponent implem
...
+	public normalizedFormButtonsConfig!: StarkGenericSearchFormButtonsConfigRequired;
 
 	/**
 	 * Reference to the FormGroup instance bound to the search form.
 	 *
 	 * In the HTML, the component defining the search form content should be exported as `searchForm`
 	 * so that it is accessible from the Stark Generic Search component. See example above.
 	 */
-	public genericForm: FormGroup;
+	public genericForm!: FormGroup;
```
christophercr: I think we should use the `?` modifier here because the `genericForm` depends on the fact that the `searchFormComponent` is present... and that one has the `?` modifier too

packages/stark-ui/src/modules/input-mask-directives/directives/email-mask.directive.spec.ts
```diff
@@ -16,8 +16,8 @@ describe("EmailMaskDirective", () => {
 		template: getTemplate("[starkEmailMask]='emailMaskConfig'")
 	})
 	class TestComponent {
-		public emailMaskConfig: boolean;
-		public ngModelValue: string = "";
+		public emailMaskConfig?: boolean;
```
christophercr: Now with this change you can remove some useless casts like `hostComponent.emailMaskConfig = <any>undefined;`

packages/stark-ui/src/modules/input-mask-directives/directives/timestamp-mask.directive.ts
```diff
@@ -55,7 +55,7 @@ export class StarkTimestampMaskDirective extends MaskedInputDirective implements
 	 */
 	/* tslint:disable:no-input-rename */
 	@Input("starkTimestampMask")
-	public maskConfig: StarkTimestampMaskConfig;
+	public maskConfig: StarkTimestampMaskConfig = { format: "DD-MM-YYYY hh:mm:ss" };
```
christophercr: Can you create a constant for the format string so we can re-use it here and also in the `private readonly defaultTimestampMaskConfig` property? Because in fact we are now using a different format here than the one over there :(

packages/stark-ui/src/modules/message-pane/components/message-pane.component.spec.ts
```diff
@@ -205,27 +204,6 @@ describe("MessagePaneComponent", () => {
...
-			expect(hostFixture).toBeDefined();
-			expect(component).toBeDefined();
-
-			expect(component.logger).not.toBeNull();
-			expect(component.logger).toBeDefined();
-			expect(component.messagePaneService).not.toBeNull();
-			expect(component.messagePaneService).toBeDefined();
-		});
-
-		it("should have default inputs", () => {
```
christophercr: I think we should keep this unit test at least to be sure that the properties are initialized with the correct value.

packages/stark-ui/src/modules/message-pane/components/message-pane.component.ts
```diff
@@ -62,47 +62,47 @@ export class StarkMessagePaneComponent extends AbstractStarkUiComponent implemen
...
 	 * Total number of messages currently shown
 	 */
-	public totalMessages: number;
+	public totalMessages = 0;
 
 	/**
 	 * Maximum level of the messages currently shown (error, warning or info).
 	 */
-	public maxLevel: StarkMessagePaneNavItem;
+	public maxLevel: StarkMessagePaneNavItem = "infos";
```
christophercr: I'm not sure if this change will cause a regression... could you double check this?

packages/stark-ui/src/modules/message-pane/services/message-pane.service.ts
```diff
@@ -67,18 +67,14 @@ export class StarkMessagePaneServiceImpl implements StarkMessagePaneService {
...
-				normalizedMessage.id = message.id;
-			}
-
-			normalizedMessage.key = message.key;
-			normalizedMessage.interpolateValues = message.interpolateValues;
-			normalizedMessage.code = message.code;
-			normalizedMessage.type = message.type;
-			normalizedMessage.priority = message.priority;
+			normalizedMessage = new StarkMessageImpl(
+				message.id,
```
christophercr: We should check that the `message.id` is not an empty string (""), if that is the case then we should pass `undefined` so the `StarkMessageImpl` can set a default id.

packages/stark-ui/src/modules/pagination/components/pagination.component.ts
christophercr: This variable can just be initialized as `{}` cause we already set the `totalItems = 0` in the `normalizePaginationConfig()` method ;)

So we just set the value in one place. What do you think?

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
christophercr: Why was this check removed?

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
christophercr: I don't think defining a default value here is a good solution because it has a side effect: an instance is registered via the `StarkProgressIndicatorService`. This means that for every instance of this directive a new config with topic "undefined" will be registered :(

packages/stark-ui/src/modules/session-ui/components/session-card/session-card.component.ts
christophercr: Why don't we initialize it as empty string "" as the other cases similar to this one?

packages/stark-ui/src/modules/slider/components/slider.component.ts
christophercr: I'm not sure if the slider will work correctly if we define this default range... Could you double check? This is the doc of this feature: https://refreshless.com/nouislider/slider-values/#section-range

packages/stark-ui/src/modules/slider/components/slider.component.ts
christophercr: I'll have a look on this and let you know if this should be changed 😉

packages/stark-ui/src/modules/table/components/table.component.spec.ts
christophercr: Not functional but more internal stuff... otherwise it will be hard to track down issues due to a wrong initialization of our properties...

packages/stark-ui/src/modules/breadcrumb/components/breadcrumb.component.ts
cnomes: 🙈 Also the reason why the build is failing I think.

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
cnomes: I looked at the code a bit better and I actually think an empty array is more correct here, or making it optional. Because it will only contain a value if `preserveLatestResults` is set to true

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
cnomes: Good catch, I initialy did this because of `this.workingCopy` not having an initial value. But later I changed it to a definite assignment and forgot to revert this code.
Although I don't know if this entire package is safe to use in it's current state. Because `this.workingCopy` is assigned in a subscription. 🤔

packages/stark-ui/src/modules/generic-search/components/generic-search/generic-search.component.ts
cnomes: I refactored this a bit to
```
	public get genericForm(): FormGroup | void {
		if (typeof this.searchFormComponent !== "undefined") {
			return this.searchFormComponent.searchForm;
		}
	}
```
but you are correct about the optional. Also, looking at the code again I think maybe I should mark `searchFormComponent` as definite and add an assert in the onInit.
following the logic of this comment https://github.com/angular/angular/issues/24571#issuecomment-404606595
What do you think?

packages/stark-ui/src/modules/input-mask-directives/directives/timestamp-mask.directive.ts
cnomes: Ah yes indeed. 🤔 I changed it to this.
```
/**
	 * Default configuration.
	 * It will be merged with the configuration passed to the directive.
	 */
	private readonly defaultTimestampMaskConfig: StarkTimestampMaskConfig = {
		format: DEFAULT_DATE_TIME_FORMAT
	};

	/**
	 * Configuration object for the mask to be displayed in the input field.
	 */
	/* tslint:disable:no-input-rename */
	@Input("starkTimestampMask")
	public maskConfig: StarkTimestampMaskConfig = this.defaultTimestampMaskConfig;
```

packages/stark-ui/src/modules/message-pane/components/message-pane.component.ts
cnomes: This gets it value through `this.getMaxLevel(msgCollection)` which returns "infos" by default. So I think this should be fine.

packages/stark-ui/src/modules/message-pane/services/message-pane.service.ts
cnomes: The `StarkMessageImpl` should cover that case through `this.id = id || _uniqueId();`

packages/stark-ui/src/modules/pagination/components/pagination.component.ts
cnomes: I don't think it will get into that logic when the paginationConfig is defined. So I would opt to remove it from the `normalizePaginationConfig` method. Or we could put it as a definite assignment and do the assignment in the `normalizePaginationConfig`.

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
cnomes: a bad fix for a build error 😅

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
christophercr: True, I think we should make it optional (`?`) then ;)

packages/stark-ui/src/modules/generic-search/classes/abstract-search-component.ts
christophercr: Well, the subscription is a pre-requisite for the GenericSearch component to work and it is documented so the developer is aware of it :wink:

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
cnomes: Maybe it's better to make it a definite assignment and add 
```
if (!this.starkProgressIndicator) {
			throw new Error("[starkProgressIndicator] is required when using the starkProgressIndicator directive ");
		}
```
to the onInit?

packages/stark-ui/src/modules/generic-search/components/generic-search/generic-search.component.ts
christophercr: Yes, I agree that `searchFormComponent` should be marked as definite and check it on the `ngOnInit` or the other hook to check child content.

So you can remove the setter you just added ;)

packages/stark-ui/src/modules/message-pane/services/message-pane.service.ts
christophercr: Ah yes, my bad, I was thinking that empty string was evaluated to `true` but it is not the case :p

packages/stark-ui/src/modules/pagination/components/pagination.component.ts
christophercr: I vote for option 2: we should mark it as definite and leave the assignment in the `normalizePaginationConfig()`

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
christophercr: Yes, maybe that solution is better 👍

packages/stark-ui/src/modules/slider/components/slider.component.ts
cnomes: I changed this to a definite with a onInit check in the meantime. 😄

packages/stark-ui/src/modules/slider/components/slider.component.ts
cnomes: I will change it to a definite with a check. 😄

cnomes:
@christophercr I updated the PR with the requested changes. 😄

packages/stark-ui/src/modules/generic-search/components/generic-search/generic-search.component.ts
```diff
@@ -235,8 +237,9 @@ export class StarkGenericSearchComponent extends AbstractStarkUiComponent implem
 	 * Component lifecycle hook
 	 */
 	public ngAfterContentInit(): void {
-		if (typeof this.searchFormComponent !== "undefined") {
-			this.genericForm = this.searchFormComponent.searchForm;
+		// TODO write a test for this
+		if (!this.searchFormComponent) {
+			throw new Error("#searchFormComponent content-child should be added to stark-generic-search");
```
christophercr: Could you add a test for this? And also, could you adapt the error message to follow the same format as the other errors: `StarkXXComponent: the error message`

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
```diff
@@ -49,13 +49,13 @@ const directiveName: string = "[starkProgressIndicator]";
 })
 export class StarkProgressIndicatorDirective implements OnInit, OnDestroy {
 	@Input()
-	public starkProgressIndicator: StarkProgressIndicatorConfig;
+	public starkProgressIndicator?: StarkProgressIndicatorConfig;
```
christophercr: I though we agreed on marking this as a definite assignment...

packages/stark-ui/src/modules/progress-indicator/directives/progress-indicator.directive.ts
```diff
@@ -88,6 +88,10 @@ export class StarkProgressIndicatorDirective implements OnInit, OnDestroy {
 	 * Finally, if the component should be hidden or shown, the stark-hide class is removed/added accordingly
 	 */
 	public ngOnInit(): void {
+		// TODO write a test for this
+		if (!this.starkProgressIndicator) {
+			throw new Error("[starkProgressIndicator] is required when using the starkProgressIndicator directive ");
```
christophercr: Can you add a unit test for this? And also, can you adapt the error message to align with the convention we follow for other errors: `StarkXXDirective: error message`.

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -110,6 +110,13 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
 	 * Component lifecycle hook that is called after data-bound properties of a directive are initialized.
 	 */
 	public ngOnInit(): void {
+		if (!this.values) {
+			throw new Error("property values should be set on stark-slider");
```
christophercr: Can you add a unit test for this? And also can you adapt the error message to align with the convention we follow for other errors: `StarkXXDirective: error message.`?

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -110,6 +110,13 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
...
+		}
+		if (!this.sliderConfig) {
+			throw new Error("property sliderConfig should be set on stark-slider");
```
christophercr: Can you add a unit test for this? And also can you adapt the error message to align with the convention we follow for other errors: `StarkXXDirective: error message.`?

showcase/src/app/welcome/pages/getting-started/getting-started-page.component.ts
```diff
@@ -14,55 +14,32 @@ import {
...
+		"# update Webdriver (optional, done automatically by postinstall script)\nnpm run webdriver:update # cfr #35\n\n# this will start a test server and launch Protractor\nnpm run e2e";
+	public continuousIntegrationTests = "# this will test both your JIT and AoT builds\nnpm run ci";
+	public installType = "npm install @types/node\nnpm install @types/lodash";
+	public moduleDeclaration = 'declare module "my-module" {\n\texport function doesSomething(value: string): string;\n}';
+	public prototyping = "declare var assert: any;\ndeclare var _: any;\ndeclare var $: any;";
+	public moduleImport = 'import * as _ from "lodash";';
 
+	public constructor(@Inject(STARK_LOGGING_SERVICE) public loggingService: StarkLoggingService) {}
+
+	public ngOnInit(): void {
```
christophercr: I think we can safely remove the `constructor` and `ngOnInit` because they are no relevant

showcase/src/app/demo-ui/pages/generic-search/components/demo-generic-search-form.component.ts
```diff
@@ -24,7 +24,7 @@ export class DemoGenericSearchFormComponent implements OnInit, OnChanges, StarkS
 	public heroOptions: string[] = [];
 	public movieOptions: string[] = [];
 
-	public searchForm: FormGroup;
+	public searchForm: FormGroup = this.createSearchForm(this.searchCriteria);
```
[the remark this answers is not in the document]
cnomes: But this logic is just for the default value/initialization, if the the input is set there is already logic in ngOnChanges to handle the update. Otherwise I'm doing the same thing twice. 😕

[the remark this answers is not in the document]
cnomes: I was wrong for the initialization, but I refactored that does follow that logic.

showcase/src/app/shared/components/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,13 +67,15 @@ describe("TableOfContents", () => {
 
 	it("should have created links", () => {
 		let links: TableOfContentLink[];
+		fixture.componentInstance.headerSelectors = "h2";
```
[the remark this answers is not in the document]
cnomes: It is because of the new logic. Otherwise the test fails.

showcase/src/app/shared/effects/stark-error-handling.effects.ts
```diff
@@ -11,7 +11,7 @@ import uniqueId from "lodash-es/uniqueId";
  */
 @Injectable()
 export class StarkErrorHandlingEffects {
-	private _starkToastNotificationService: StarkToastNotificationService;
+	private _starkToastNotificationService?: StarkToastNotificationService;
```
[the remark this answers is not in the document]
cnomes: But we do the `if (typeof this._starkToastNotificationService === "undefined")`, so that does mean we expect it to not be defined in some cases right? 🤔

cnomes:
@christophercr I updated the PR and asked some questions on some Change requests. 😄

showcase/src/app/shared/effects/stark-error-handling.effects.ts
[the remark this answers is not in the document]
christophercr: Indeed, but that is just the first time when the service has not been injected... then the service will always be defined.

But ok... either `?` or `!` are fine for me... this doesn't affect the logic in any case ;)

showcase/src/app/demo-ui/pages/generic-search/components/demo-generic-search-form.component.ts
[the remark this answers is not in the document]
cnomes: Add back the isEqual

cnomes:
PR updated and build passing. 💪

cnomes:
I also updated the commit messages / PR description and updated `stark-rbac`

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -110,8 +110,18 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
 	 * Component lifecycle hook that is called after data-bound properties of a directive are initialized.
 	 */
 	public ngOnInit(): void {
-		this.logger.debug(componentName + ": component initialized");
+		console.log("BEFORE CHECKS:", this.values);
```
christophercr: Can you remove this `console.log`?

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -110,8 +110,18 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
...
+		if (!this.values) {
+			throw new Error("StarkSliderComponent: values should be set.");
+		}
+		console.log("AFTER VALUES CHECK: ", this.values);
```
christophercr: Can you remove this `console.log`?

packages/stark-ui/src/modules/slider/components/slider.component.ts
```diff
@@ -110,8 +110,18 @@ export class StarkSliderComponent extends AbstractStarkUiComponent implements Af
...
+		if (!this.sliderConfig) {
+			throw new Error("StarkSliderComponent: sliderConfig should be set.");
+		}
+		console.log("AFTER CONFIG CHECK: ", this.sliderConfig);
```
christophercr: Can you remove this `console.log`?

showcase/src/assets/examples/transform-input-directive/transform-input-directive.ts
```diff
@@ -1,27 +1,27 @@
...
 
 @Component({
 	selector: "demo-transform-input",
 	templateUrl: "./demo-transform-input.component.html"
 })
-export class DemoTransformInputComponent implements OnInit {
-	public value: string = "";
-	public formControl: FormControl = new FormControl("");
-	private _emojiMap: Map<string, string> = new Map([
+export class DemoTransformInputComponent {
```
christophercr: Can you remove the `translate` directive from the HTML file of this example?

showcase/src/assets/examples/app-data/dropdown.scss
```diff
@@ -0,0 +1,18 @@
+.dropdown-app-data.with-user-image {
```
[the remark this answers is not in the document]
cnomes: It is referenced, but only in the last example. I changed the `showcase/src/app/demo-ui/pages/app-data/_demo-app-data-page.component.scss`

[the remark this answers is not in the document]
cnomes: It is referenced, but only in the first example

showcase/src/app/demo-ui/pages/collapsible/demo-collapsible-page.component.html
```diff
@@ -2,7 +2,7 @@ <h1 class="mat-display-3" translate>SHOWCASE.DEMO.COLLAPSIBLE.TITLE</h1>
 <section class="stark-section">
 	<h1 translate>SHOWCASE.DEMO.SHARED.EXAMPLE_VIEWER_LIST</h1>
 	<example-viewer
-		[extensions]="['HTML', 'CSS', 'TS']"
+		[extensions]="['HTML', 'SCSS', 'TS']"
```
christophercr: In fact, looking at the SCSS code, I think we can simply remove this file: `showcase/src/assets/examples/collapsible/collapsible-default.scss` because those styles are not relevant for the example

showcase/src/app/demo-ui/pages/collapsible/demo-collapsible-page.component.html
```diff
@@ -18,7 +18,7 @@ <h1 translate>SHOWCASE.DEMO.SHARED.EXAMPLE_VIEWER_LIST</h1>
 	</example-viewer>
 
 	<example-viewer
-		[extensions]="['HTML', 'CSS', 'TS']"
+		[extensions]="['HTML', 'SCSS', 'TS']"
```
christophercr: Same here, looking at the SCSS code, I think we can simply remove this file: `showcase/src/assets/examples/collapsible/collapsible-custom.scss` because those styles are not relevant for the example

showcase/src/assets/examples/app-data/dropdown.ts
```diff
@@ -1,12 +1,33 @@
...
+	StarkApplicationMetadata,
+	StarkUser,
+	StarkUserService
+} from "@nationalbankbelgium/stark-core";
+import * as moment from "moment";
+import { filter } from "rxjs/operators";
 
 @Component({
 	selector: "demo-app-data",
 	templateUrl: "./demo-app-data.component.html"
```
christophercr: Can you add the `styleUrls` here? Because this example also includes some CSS ;)

showcase/src/assets/examples/generic-search/demo-generic-search-page.component.ts
```diff
@@ -1,81 +1,154 @@
...
-	templateUrl: "./demo-generic-search-page.component.html"
+	selector: "demo-message-pane",
+	templateUrl: "./demo-message-pane-page.component.html"
 })
-export class DemoGenericSearchPageComponent extends AbstractStarkSearchComponent<HeroMovie, HeroMovieSearchCriteria>
-	implements OnInit, OnDestroy {
-	public columnsProperties: StarkTableColumnProperties[];
-	public searchResults: HeroMovie[];
-	public paginationConfig: StarkPaginationConfig;
+export class DemoMessagePanePageComponent {
```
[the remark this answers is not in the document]
cnomes: 😅

cnomes:
@christophercr Updated. 😉

christophercr:
@SuperITMan I've reviewed all the modules in this PR (build, core, ui, rbac, starter and showcase) so I think we can merge this. What do you think?

packages/stark-core/src/modules/user/services/user.service.ts
```diff
@@ -79,7 +78,7 @@ export class StarkUserServiceImpl implements StarkUserService {
 
 		const allUsers: StarkUser[] = this.userProfiles;
 
-		if (!allUsers || allUsers.length === 0) {
+		if (allUsers.length === 0) {
```
SuperITMan: We should keep the condition as before :
```js
if (!allUsers || allUsers.length === 0)
```

Because when we deserialize the object in the constructor, we are not sure that the deserialization will be fine.
It could return `null` or `undefined` and so we'll have the error `cannot read length of undefined`...

christophercr: I've updated this to fallback to `[]` in case the deserialization returns `null` or `undefined`... so we always return something ;)

showcase/src/app/shared/components/table-of-contents/table-of-contents.component.ts
```diff
@@ -19,11 +19,11 @@ export class TableOfContentsComponent implements OnInit, AfterViewInit, OnDestro
 	 * The classes of the titles you want to use in your table of contents
 	 */
 	@Input()
-	public headerSelectors: string;
+	public headerSelectors?: string;
 
-	private _scrollContainer: Window;
+	private _scrollContainer: Window = window;
```
SuperITMan: If we set the `_scrollContainer` to `window` here, why doing it again at line 39 in the `ngOnInit` ?
I think we should remove this assignment...

christophercr: indeed, that's not needed anymore. I'll fix it 👍

christophercr:
PR updated with the latest remarks :wink:

PR #1233, opened 2019-04-02
docs(stark-build): update urls / comments in `config/webpack-partialcommon.js`

cnomes:


packages/stark-build/config/webpack-partial.common.js
```diff
@@ -65,22 +65,19 @@ module.exports = metadata => {
 		/**
 		 * Options affecting the normal modules.
 		 *
-		 * See: http://webpack.github.io/docs/configuration.html#module
+		 * See: https://webpack.js.org/configuration/module
 		 */
 		module: {
 			rules: [
-				// TsLint loader support for *.ts files
-				// reference: https://github.com/wbuchwalter/tslint-loader
```
christophercr: Can we keep these 2 lines at least? So we have a reference on where to find the docs

cnomes: Updated

cnomes:
I'm merging this with #1238 and continuing there.

Issue #956, opened 2018-12-17
core: user profile - make the base path configurable for the User Profile end point in the User Repository


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

The base path of the user profile end point is hard-coded in User Repository to `security/userprofile` which might not be always the case for some applications.

## Expected behavior

The base path should be configurable so the developer can adapt it according to his needs.

## Minimal reproduction of the problem with instructions

1. In your backend configure a User Profile end point with a different URL other than `security/userprofile`.
2. The target URL of the Http call to fetch the user profile will still point to `security/userprofile`.

## What is the motivation / use case for changing the behavior?

Provide a flexible and easy to configure core API.

## Environment

<pre><code>
Angular version: 7.1.0
Stark version: 10.0.0-beta.1

</code></pre>

PR #1250, opened 2019-04-12
feat(stark-core): allow configuring the User profile resource path used by the User Repository

christophercr:
ISSUES CLOSED: #956

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #956 

## What is the new behavior?
- a config object can be passed to the User Module to set the User profile resource path to be used by the User Repository
- a new `STARK_USER_PROFILE_RESOURCE_PATH` injection token is created to pass the custom resource path along to the User Repository
- implement missing unit tests for the User Repository

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-core/src/modules/user/entities/user-module-config.entity.intf.ts
+++ b/packages/stark-core/src/modules/user/entities/user-module-config.entity.intf.ts
@@ -1,0 +1,18 @@
+import { InjectionToken } from "@angular/core";
+
+/**
+ * The InjectionToken version of the config value
+ */
+export const STARK_USER_PROFILE_RESOURCE_PATH = new InjectionToken<StarkUserModuleConfig["userProfileResourcePath"]>(
+	"StarkUserProfileResourcePath"
+);
+
+/**
+ * Definition of the configuration object for the Stark User Module
+ */
+export interface StarkUserModuleConfig {
+	/**
+	 * Base path to be used in the Http requests to fetch the user profile resource
+	 */
+	userProfileResourcePath: string;
+}
--- a/packages/stark-core/src/modules/user/user.module.ts
+++ b/packages/stark-core/src/modules/user/user.module.ts
@@ -1,4 +1,5 @@
 import { ModuleWithProviders, NgModule, Optional, SkipSelf } from "@angular/core";
+import { STARK_USER_PROFILE_RESOURCE_PATH, StarkUserModuleConfig } from "./entities";
 import { STARK_USER_SERVICE, StarkUserServiceImpl } from "./services";
 import { STARK_USER_REPOSITORY, StarkUserRepositoryImpl } from "./repository";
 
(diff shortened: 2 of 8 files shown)


packages/stark-core/src/modules/user/user.module.ts
```diff
@@ -8,14 +9,16 @@ export class StarkUserModule {
...
 	 */
-	public static forRoot(): ModuleWithProviders {
+	public static forRoot(userModuleConfig?: StarkUserModuleConfig): ModuleWithProviders {
 		return {
 			ngModule: StarkUserModule,
 			providers: [
 				{ provide: STARK_USER_SERVICE, useClass: StarkUserServiceImpl },
-				{ provide: STARK_USER_REPOSITORY, useClass: StarkUserRepositoryImpl }
+				{ provide: STARK_USER_REPOSITORY, useClass: StarkUserRepositoryImpl },
+				userModuleConfig ? { provide: STARK_USER_PROFILE_RESOURCE_PATH, useValue: userModuleConfig.userProfileResourcePath } : []
```
SuperITMan: This code seems weird to me...
Should not it be: 

```typescript
providers: [
  { provide: STARK_USER_SERVICE, useClass: StarkUserServiceImpl },
  { provide: STARK_USER_REPOSITORY, useClass: StarkUserRepositoryImpl },
  ... userModuleConfig ? { provide: STARK_USER_PROFILE_RESOURCE_PATH, useValue: userModuleConfig.userProfileResourcePath } : []
]
```

instead ? :blush:

Otherwise, if there is no `userModuleConfig` you will have the following result object which contains an empty array : 
```typescript
providers: [
  { provide: STARK_USER_SERVICE, useClass: StarkUserServiceImpl },
  { provide: STARK_USER_REPOSITORY, useClass: StarkUserRepositoryImpl },
  []
]
```

christophercr: About your suggestion... I don't understand... what should be changed?

In fact, `[]` is a valid value to pass to module `providers`... this is the typings for `Provider`:

```typescript
export declare type Provider = TypeProvider | ValueProvider | ClassProvider | ConstructorProvider | ExistingProvider | FactoryProvider | any[];
```

christophercr: What I can do is to change it like this:

```typescript
userModuleConfig ? { provide: STARK_USER_PROFILE_RESOURCE_PATH, useValue: userModuleConfig.userProfileResourcePath } : ...[]
```

so I expand the array `...[]` insteadof passing `[]`

SuperITMan: I think it's weird to pass an empty array to providers. 
Your proposal seems fine since the final `providers` won't contain a useless empty array 😊

christophercr: @SuperITMan I'm afraid I cannot change the code as we want because using the spread operator `...` or any function causes the AOT compilation to fail: `Function calls are not supported in decorators but 'StarkUserModule' was called.`.

This is because the spread operator is transpiled to `Array.concat()` in ES5.... and functions cannot be called in AOT :(

So I'll leave the code as it is now, in fact we have done this in other modules... and it is the way is recommended in forums, for example: https://stackoverflow.com/a/49763160

packages/stark-core/src/modules/user/entities/user-module-config.entity.intf.ts
```diff
@@ -0,0 +1,18 @@
+import { InjectionToken } from "@angular/core";
+
+/**
+ * The InjectionToken version of the config value
+ */
+export const STARK_USER_PROFILE_RESOURCE_PATH: InjectionToken<StarkUserModuleConfig["userProfileResourcePath"]> = new InjectionToken<
```
cnomes: I don't really think the type is necessary here, what do you think?

cnomes: I don't really think the type is necessary here, what do you think?

PR #1258, opened 2019-04-17
feat(stark-ui): implement `ControlValueAccessor` on `DateRangePicker` component

cnomes:
- provide functionality for displaying errors
  - added example
  - added test

ISSUES CLOSED: #1197

BREAKING CHANGES: `stark-date-range-picker` is now implemented as a ControlValueAccessor to better integrate with Angular Forms:

  - new `rangeFormGroup` input to provide the start and end date form controls. You can handle validations and error messages via those form controls.
  - new `required` input to mark the control as required and display the required asterisk '*' in the start/end date pickers (same as with the `stark-date-picker`).
  - the `isDisabled` input has been renamed to `disabled`. If you use the new `rangeFormGroup` input to provide your start/end date form controls, then you should disable/enable those form controls directly yourself instead of using the `disabled` input.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1197 

## What is the new behavior?
StarkDateRangePicker now works with ngModel and FormControl

```
[X] Yes (`isDisabled`is now `disabled`)
[ ] No
```


cnomes:
@christophercr I updated this PR

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +38,151 @@ const componentName = "stark-date-range-picker";
...
-	 * Maximum date of the end date picker
+	 * Reference to the end datepicker embedded in this component
 	 */
-	@Input()
-	public endMaxDate?: Date;
+	@ViewChild("endPicker")
+	public endPicker!: StarkDatePickerComponent;
+
+	/*--- START-DATE CONFIGURATIONS ---*/
+	private _startBeforeEndVal: ValidatorFn = ({ value }) => {
```
christophercr: Can you rename this to `_startBeforeEndValidator` to make it clear it is a validator?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +38,151 @@ const componentName = "stark-date-range-picker";
...
 	/**
-	 * Minimum date of the start date picker
+	 * Input to manage both start date and end date.
 	 */
 	@Input()
-	public startMinDate?: Date;
+	public set formGroup(val: FormGroup) {
+		const { startDate, endDate } = val.controls;
+		if (!(startDate instanceof FormControl && endDate instanceof FormControl)) {
+			this.logger.warn(`[${componentName}]: formGroup requires a FormControl for startDate and endDate`);
```
christophercr: Shouldn't this be an error? Otherwise the date-range-picker won't work right?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +191,140 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
+		// Get the ngControl from the injector
+		try {
+			const ngControl = this.injector.get<NgControl>(<Type<NgControl>>NgControl);
+			ngControl.valueAccessor = this;
+
+			// ngControl.control is not available at init
+			setTimeout(
+				(): void => {
+					if (typeof get(ngControl, "control.validator") === "function") {
+						this.logger.warn(`[${componentName}]: use formGroup for custom validators`);
```
christophercr: Maybe this warning should be more explanatory... what happens if they don't use formGroup?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +191,140 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
 
 	/**
-	 * Handle the date changed on the end datepicker
+	 * Angular lifecycle method
 	 */
-	public onDateEndChanged(date: Date): void {
-		this.endDate = date;
-		this.checkDates();
+	public ngOnDestroy(): void {
+		this.subs.forEach((s: Subscription) => s.unsubscribe());
```
christophercr: Just a suggestion: a `for...of` loop is more compact

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +38,151 @@ const componentName = "stark-date-range-picker";
...
-	 * HTML "name" attribute of the element.
+	 * Maximum date of the start date picker
 	 */
 	@Input()
-	public rangePickerName = "";
+	public startMaxDate?: Date;
+
+	/*--- END-DATE CONFIGURATIONS ---*/
+
+	private _endAfterStartVal: ValidatorFn = ({ value }) => {
```
christophercr: Can you rename this to `_endAfterStartValidator `to make it clear it is a validator?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +191,140 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
+	 * @internal
+	 * @ignore
 	 */
-	public onDateStartChanged(date: Date): void {
-		this.startDate = date;
-		this.checkDates();
+	private _setupNgControl(): void {
+		// Get the ngControl from the injector
+		try {
+			const ngControl = this.injector.get<NgControl>(<Type<NgControl>>NgControl);
```
christophercr: Here, you can simplify the code to avoid having a `try...catch` block. Something like this:

```typescript
const ngControl = this.injector.get<NgControl>(<Type<NgControl>>NgControl, <any>null);  // default value: null

if (ngControl !== null) {
    ngControl.valueAccessor = this;
}
```

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +191,140 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
+			ngControl.valueAccessor = this;
+
+			// ngControl.control is not available at init
+			setTimeout(
```
christophercr: Can this `setTimeout` be replaced by putting the logic in the `ngAfterViewInit()` hook?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +191,140 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
-	public endPicker!: StarkDatePickerComponent;
+	@Input()
+	public dateMask?: StarkDatePickerMaskConfig;
+
+	/**
+	 * Whether the datepickers are disabled
+	 */
+	@Input()
+	public set disabled(val: boolean) {
+		if (val) {
```
christophercr: I think we should log a warning here in case the developer is using the `formGroup` input and also this `disabled` input... which might cause some confusion and prone to errors on his side. So we could warn him about this so if he uses the `formGroup`, he can simply disable the group. What do you think?

showcase/src/assets/translations/fr.json
```diff
@@ -52,9 +52,12 @@
         "TITLE": "Date picker"
       },
       "DATE_RANGE_PICKER": {
-        "CUSTOM_MASK": "Date range picker - Masque de date personnalisé ({{ mask }})",
-        "DEFAULT_MASK": "Date range picker - Masque de date par défaut",
-        "TITLE": "Date range picker"
+        "TITLE": "Date range picker",
+        "WITH_MODEL": "Angular model",
```
christophercr: I think "NgModel" sounds more familiar than "Angular Model"

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
```diff
@@ -1,80 +1,57 @@
...
+			dateFilter="OnlyWeekdays"
+			[startMinDate]="today"
+			[endMinDate]="today"
+			[startMaxDate]="inOneMonth"
+			[endMaxDate]="inOneMonth"
+			[disabled]="modelDisabled"
+		></stark-date-range-picker>
+		<br />
+		<mat-checkbox [(ngModel)]="modelDisabled">
+			{{ (modelDisabled ? "SHOWCASE.COMMON.DISABLED" : "SHOWCASE.COMMON.ENABLED") | translate }}
```
christophercr: Can you add the same functionality for the other example with FormGroup? In that case it is just about setting the formGroup as disabled right?

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.ts
```diff
@@ -1,30 +1,38 @@
...
-	public customDateFilter: StarkDatePickerFilter = (date: Date): boolean => date.getDay() !== 0;
+	public dateRangeModel = { startDate: this.today, endDate: this.inOneMonth };
+	public modelDisabled = false;
 
-	public customDateMask: StarkTimestampMaskConfig = {
-		format: "DD-MM-YYYY"
-	};
+	public dateRangeFormGroup = new FormGroup({
+		startDate: new FormControl(null, DemoDateRangePickerPageComponent.noFebruaryValidator),
+		endDate: new FormControl(null, DemoDateRangePickerPageComponent.noFebruaryValidator)
```
christophercr: Can you add also a `required` validator to show that it is correctly handled?

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
```diff
@@ -1,80 +1,57 @@
...
+		exampleTitle="SHOWCASE.DEMO.DATE_RANGE_PICKER.WITH_MODEL"
+	>
+		<stark-date-range-picker
+			[(ngModel)]="dateRangeModel"
+			(dateRangeChanged)="onDateModelChange()"
+			dateFilter="OnlyWeekdays"
+			[startMinDate]="today"
+			[endMinDate]="today"
+			[startMaxDate]="inOneMonth"
+			[endMaxDate]="inOneMonth"
```
christophercr: In this example the errors for min and max dates are not displayed:

date-range-picker no errors

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.ts
```diff
@@ -33,12 +41,19 @@ export class DemoDateRangePickerPageComponent {
...
 
-	public constructor(@Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService) {}
+	public constructor(@Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService) {
+		this._subs.push(this.dateRangeFormGroup.valueChanges.subscribe((v: any) => this.logger.debug(v)));
+	}
 
-	public onDateChanged(event: StarkDateRangePickerEvent): void {
-		this.logger.debug(event);
+	public onDateModelChange(): void {
+		this.logger.debug(this.dateRangeModel);
```
christophercr: Can you add some text to this logging to know what is being logged? Otherwise is confusing in the browser console...

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.ts
```diff
@@ -33,12 +41,19 @@ export class DemoDateRangePickerPageComponent {
...
+	}
+
+	public onDateChange(dateRange: StarkDateRangePickerEvent): void {
+		this.logger.debug(dateRange);
```
christophercr: Can you add some text to this logging to know what is being logged? Otherwise is confusing in the browser console...

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
christophercr: Currently in the Showcase, the message for this error is not defined

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
christophercr: Currently in the Showcase, the message for this error is not defined

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
cnomes: It will still work, but their form wont manage the stark-date-picker

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
cnomes: is "validators set on the control will not be used, use the formGroup input to manage your own validation."  clearer?

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
cnomes: Indeed, there are no error messages defined by default for any validation. Should there be?

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
cnomes: It's weird, but normally you can use both. I added a TMP commit to demonstrate

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
christophercr: Not by default... I mean, they should not be defined in Stark but in the demo page itself ;)

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
christophercr: Mmm, well, throwing an error should be fine... let's make it as painful as possible :p

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
christophercr: Yeah, that looks better... just put the "formGroup" between quotes ('') and also "validations" in plural

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
christophercr: Something similar to what we did in `src/app/demo-ui/pages/date-picker/demo-date-picker-page.component.html`

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
christophercr: Yes, exactly. It is a weird scenario but yet is possible. So we should at least warn the developer that he is providing 2 inputs which are not intended to be used together...

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
cnomes: 😂 👍

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
cnomes: I updated the demo with the messages

cnomes:
I updated the demo template to fix the `cyclomatic complexity`

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +39,155 @@ const componentName = "stark-date-range-picker";
...
-	public endMinDate?: Date;
+	public get startDate(): Date | undefined {
+		return this._startDate.value || undefined;
+	}
+
+	public set startDate(value: Date | undefined) {
+		this._startDate.setValue(value);
+	}
+
+	public get startDateFormControl(): FormControl {
```
christophercr: We need to add JSDoc for this because I see that in the Demo with NgModel we do something like this: `*ngFor="let error of getErrorMessages(dateRangeComponent.startDateFormControl)"`. So they should be included in the docs ;)

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +39,155 @@ const componentName = "stark-date-range-picker";
...
+	@Input()
+	public get endDate(): Date | undefined {
+		return this._endDate.value || undefined;
+	}
+
+	public set endDate(value: Date | undefined) {
+		this._endDate.setValue(value || null);
+	}
+
+	public get endDateFormControl(): FormControl {
```
christophercr: We need to add JSDoc for this because I see that in the Demo with NgModel we do something like this: `*ngFor="let error of getErrorMessages(dateRangeComponent.endDateFormControl)"`. So they should be included in the docs ;)

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.ts
```diff
@@ -1,30 +1,66 @@
...
+	public dateRangeFormGroup = new FormGroup({
+		startDate: new FormControl(null, Validators.compose([DemoDateRangePickerPageComponent.noFebruaryValidator, Validators.required])),
+		endDate: new FormControl(null, Validators.compose([DemoDateRangePickerPageComponent.noFebruaryValidator, Validators.required]))
+	});
 
-	public customDateMask: StarkTimestampMaskConfig = {
-		format: "DD-MM-YYYY"
-	};
+	public getErrorMessages(control: AbstractControl): string[] {
+		return map(
```
christophercr: Question:  what is the added value of using Lodash here instead of simply Array.map() ?

cnomes: control.errors is an object, so I'm not able to use the Array.map(). But I should change the fallback 😅

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +39,185 @@ const componentName = "stark-date-range-picker";
...
 	/**
-	 * Source Date to be bound to the start datepicker model
+	 * The internal formControl used to manage the state of the starkDatePicker for the start date.
+	 * This can be used to retrieve internal errors.
+	 *
+	 * @example
+	 *
+	 * <stark-date-range-picker required #starkDateRangePicker >
+	 *   <ng-container start-date-errors>
+	 *     {{starkDateRangePicker.startDateFormControl.hasError('required') ? "Start date is required" : null }}
```
christophercr: This should be `starkDateRangePicker.startDateFormControl` instead 😉

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -117,67 +226,175 @@ export class StarkDateRangePickerComponent extends AbstractStarkUiComponent impl
...
+			this.startDateFormControl.disable();
+			this.endDateFormControl.disable();
+		} else {
+			this.startDateFormControl.enable();
+			this.endDateFormControl.enable();
+		}
+	}
+
+	/**
+	 * Wheter the datepickers are required
```
christophercr: Typo: "Whether"

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
```diff
@@ -1,80 +1,79 @@
...
+			{{ (modelDisabled ? "SHOWCASE.COMMON.DISABLED" : "SHOWCASE.COMMON.ENABLED") | translate }}
+		</mat-checkbox>
+	</example-viewer>
+
+	<example-viewer
+		[extensions]="['HTML', 'TS']"
+		filesPath="date-range-picker/form-group"
+		exampleTitle="SHOWCASE.DEMO.DATE_RANGE_PICKER.WITH_FORM_GROUP"
+	>
+		<stark-date-range-picker [formGroup]="dateRangeFormGroup">
```
christophercr: The `required` attribute is missing in this example...  because the formControls do have the `Validators.required`

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +39,185 @@ const componentName = "stark-date-range-picker";
...
 
 	/**
-	 * Source Date to be bound to the start datepicker model
+	 * The internal formControl used to manage the state of the starkDatePicker for the start date.
+	 * This can be used to retrieve internal errors.
+	 *
+	 * @example
+	 *
+	 * <end-date-range-picker required #endDateRangePicker >
+	 *   <ng-container start-date-errors>
```
christophercr: This line should be: `<ng-container end-date-errors>`

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
```diff
@@ -24,91 +39,185 @@ const componentName = "stark-date-range-picker";
...
+	 *     {{endDateRangePicker.endDateFormControl.hasError('required') ? "End date is required" : null }}
```
christophercr: This line should be: `starkDateRangePicker.endDateFormControl.hasError....`

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
```diff
@@ -1,80 +1,80 @@
...
```
christophercr: The `required` attribute is missing in this example...  because the formControls do have the `Validators.required`.

The corresponding example file should be changed as well 😉

packages/stark-ui/src/modules/date-range-picker/components/date-range-picker.component.ts
cnomes: 🤦‍♂

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.html
cnomes: Do you mean in `showcase/src/assets/examples/date-range-picker/form-group.ts`? because the required is still used in this example. The validation just happens in the demo, not the component.

Or maybe this is to confusing and it should be removed. 🤔

christophercr:
@SuperITMan I've updated this PR to fix the unit tests after the last change from @cnomes. I also added the 'breaking changes' section to the commit message.

Could you please have a look?

showcase/src/app/demo-ui/pages/date-range-picker/demo-date-range-picker-page.component.ts
```diff
@@ -33,12 +71,31 @@ export class DemoDateRangePickerPageComponent {
...
+	}
+
+	public onDateChange(dateRange: StarkDateRangePickerEvent): void {
+		this.logger.debug("onChange:", dateRange);
+	}
 
-		this.startDate = event.startDate;
-		this.endDate = event.endDate;
+	public ngOnDestroy(): void {
+		this._subs.forEach((sub: Subscription) => sub.unsubscribe());
```
SuperITMan: Could we switch this `forEach` for a `for of` which is more efficient ? 😊

showcase/src/assets/examples/date-range-picker/form-group.ts
```diff
@@ -0,0 +1,64 @@
...
+	public onDateRangeFormGroupDisableCheckboxChange(event: MatCheckboxChange): void {
+		if (event.checked) {
+			this.dateRangeFormGroup.disable();
+		} else {
+			this.dateRangeFormGroup.enable();
+		}
+	}
+
+	public ngOnDestroy(): void {
+		this._subs.forEach((sub: Subscription) => sub.unsubscribe());
```
SuperITMan: same here about the `forEach`

christophercr:
@SuperITMan I've updated this PR with your remarks. Can you review them?

SuperITMan:
I find there is something weird in our current implementation.
When we have a range date picker with a start and an end date, we can be sure the end date should be after the start date.
What would you think, at least when there is no endMinDate and/or no startMaxDate, to set dynamically the endMinDate as the current value of startDate and the startMaxDate as the current value of endDate ? :blush:

cnomes:
@SuperITMan Sounds good to me, but there **might** be a problem with circular logic with how this component is written. But I don't remember exactly.

christophercr:
@SuperITMan @cnomes I've updated this PR with the implementation of the feature suggested in the last comment: set the dynamic endMinDate based on what the user selects in the start date.

It took me a while but finally it seems is working correctly and without the f*** `ExpressionChangedAfterItHasBeenCheckedError`  😆

christophercr:
@cnomes @SuperITMan can you guys have a look on this PR? I've done all the fixes mentioned in your remarks. This way we can release today or tomorrow the latest 😉

cnomes:
LGTM 👍

SuperITMan:
A small remark about the date range picker for usability.
Currently, if I select a **start** date in the past, when I want to select the **end** date, I have to manually go in the past to select a date closed to my **start** date.
image

I think we should use the **Input** `startAt` of the angular datepicker and set it for the **end** date to `startDate || today`. Then when you select a **start** date in the past, you can easily select an **end** date close to it.

What do you think about this @cnomes @christophercr ? 😊 
Maybe we can create an issue for this to solve in **10.0.0-rc.0**

christophercr:
Good point. I think we can tackle that in **10.0.0-rc.0** indeed. However I think we should discuss it thoroughly because that behavior really depends on the use case. But ok, we should create an issue for this and start the discussion over there 😉

SuperITMan:
Okido! 😊 
Issue created: #1314

christophercr:
Great! So let's merge it!!!! 🎉

Issue #1257, opened 2019-04-17
ui: date-picker - the invalid status is shown even if the picker is pristine


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When a date picker is initialized as empty and with the `required` validator, the picker is shown as invalid after being initialized.

## Expected behavior

The date picker should be shown as invalid only when the user has interacted with the picker or he has changed its value.

## Minimal reproduction of the problem with instructions

- Add a `stark-date-picker` element to your template.
- Provide a `FormControl` to the date picker. Initialize the form control with `undefined` and add the `required` validator to it.
- Run the application. The date picker will be immediately shown as invalid and with the "required" error message.

## What is the motivation / use case for changing the behavior?

Better UX.

## Environment

<pre><code>
Angular version: X.Y.Z
Stark version: 10.0.0-beta.6, 10.0.0-beta.7

</code></pre>

PR #1259, opened 2019-04-18
Bugfix/date picker invalid status on initialization + required asterisk not shown

### LINKED ISSUE #1255: ui: date-picker - asterisk (*) is not shown when marked as required
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When a StarkDatePicker is marked as `required`, there is no asterisk next to the label to indicate this.

## Expected behavior

A field marked as `required` should have an asterisk next to its label: https://material.angular.io/components/form-field/overview#error-messages

## Minimal reproduction of the problem with instructions

See https://stark.nbb.be/showcase/latest/demo-ui/date-picker

## What is the motivation / use case for changing the behavior?

Align with material design guidelines: https://material.io/design/components/text-fields.html#anatomy

## Environment

<pre><code>
Stark version: 10.0.0-beta.7

</code></pre>

christophercr:
ISSUES CLOSED #1255, #1257

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1255, #1257

## What is the new behavior?
- the asterisk (*) is appended to the date picker label in case it is marked as required
- in case the date picker is initialized with an invalid date, it will only be displayed as invalid until the user interacts with the picker.
- refactored the unit tests to make them more readable

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
+++ b/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
@@ -407,6 +407,10 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
 			);
 		}
 
+		if (changes["required"]) {
+			this.required = coerceBooleanProperty(changes["required"].currentValue);
+		}
+
 		if (changes["max"] || changes["min"] || changes["required"]) {
 			this.stateChanges.next();
 			this.cdRef.detectChanges();
--- a/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
+++ b/packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
@@ -1,4 +1,5 @@
 import {
+	AfterViewInit,
 	ChangeDetectorRef,
 	Component,
 	ElementRef,
@@ -78,15 +79,32 @@ const componentName = "stark-date-picker";
 	]
 })
 export class StarkDatePickerComponent extends AbstractStarkUiComponent
-	implements OnInit, OnChanges, OnDestroy, ControlValueAccessor, Validator, MatFormFieldControl<Date> {
+	implements OnInit, AfterViewInit, OnChanges, OnDestroy, ControlValueAccessor, Validator, MatFormFieldControl<Date> {
+	/**
+	 * Part of {@link MatFormFieldControl} API
+	 * @ignore
+	 * @internal
+	 */
 	public static nextId = 0;
 
+	/**
+	 * Part of {@link MatFormFieldControl} API
+	 * @ignore
+	 */
 	@HostBinding()
 	public id = `stark-date-picker-input-${StarkDatePickerComponent.nextId++}`;
 
+	/**
+	 * Part of {@link MatFormFieldControl} API
+	 * @ignore
+	 */
 	@HostBinding("attr.aria-describedby")
 	public describedBy = "";
 
+	/**
+	 * Part of {@link MatFormFieldControl} API
+	 * @ignore
+	 */
 	@HostBinding("class.floating")
 	public get shouldLabelFloat(): boolean {
 		return this.focused || !this.empty;
@@ -144,8 +162,8 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
 	public min?: Date;
 
 	/**
-	 * id attribute of the form field wrapping the mat-datepicker
-	 * id attribute followed by "-input" of the mat-datepicker-input
+	 * The HTML "id" attribute of the date picker's calendar popup.
+	 * This "id" is also used, suffixed with "-input", as the HTML "id" attribute of the date picker's input field.
 	 */
 	@Input()
 	public pickerId = "";
@@ -229,6 +247,7 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
 	public dateMaskConfig?: StarkTimestampMaskConfig = undefined;
 
 	/**
+	 * Part of {@link MatFormFieldControl} API
 	 * @ignore
 	 * @internal
 	 */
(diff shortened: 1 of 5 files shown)


packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -229,40 +247,52 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
...
+	 * Part of {@link MatFormFieldControl} API
+	 * @ignore
+	 * @internal
 	 */
 	public get errorState(): boolean {
+		// the control can be in an error state as long as the user has interacted with it
 		return (
 			this.ngControl !== null &&
 			this.ngControl.control !== null &&
+			this.pickerInputTouched &&
```
SuperITMan: Now, we are checking if it is **touched** which is better than before.
I think we should also check if it is **dirty** or not. Adding this condition would be probably better:
`(this.pickerInputTouched || this.pickerInputDirty) &&`

Othewise it won't be marked as error when the user submit the form...

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -391,6 +391,10 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
 		}
 
 		if (changes["max"] || changes["min"] || changes["required"]) {
+			if (changes["required"]) {
```
cnomes: Personally I would put this in a setter or not nest it, but it's not a blocking issue for me.

cnomes: Personally I would put this in a setter or not nest it, but it's not a blocking issue for me.

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
christophercr: I agree... but there is a problem though, submitting the form does not mark the control as `dirty`.

So in this case, it is the developer that should mark the control as `dirty` or `touched`. Moreover there is no way (in the Angular API nor Material API) to know when then control has changed to a `dirty` state so we need to write some logic to know when that happens.

I'll try to implement this and see how far I get...

SuperITMan: Ok, it's sad... Anyway, they propose to use this method in the submit method:
```typescript
/**
   * Marks all controls in a form group as touched
   * @param formGroup - The form group to touch
   */
  private markFormGroupTouched(formGroup: FormGroup) {
    (<any>Object).values(formGroup.controls).forEach(control => {
      control.markAsTouched();

      if (control.controls) {
        this.markFormGroupTouched(control);
      }
    });
  }
```

You can see it here: https://github.com/angular/angular/issues/12281 

Because of this, I think that we should change this new line by 

```js
(this.pickerInputTouched || this.ngControl.dirty) &&
```

Like this, when the user clicks on "submit" button, the developer can simply mark the control as "dirty" and it will work correctly.
What do you think ? 😊

christophercr: Yes, indeed. In fact, the developer can mark the form control as `dirty` or `touched`, that is not the issue... the problem is to know when the control has been marked as `dirty` or `touched` so we can display the errors properly. Because knowing that the formFieldControl is in an "error state" is one thing and to display the errors is another thing.

I think I have an idea... I'll try to implement it

SuperITMan: If we check if it is "touched" (via `pickerInputTouched` variable) or check if the formControl is marked as touched (via `ngControl.touched`), the MatFormField will be in error state and the message will be displayed if it is defined in the mat-error if the the dev trigger the `updateValueAndValidity()` on the FormControl...

That means, we should adapt the method as this:
```typescript
/**
   * Marks all controls in a form group as touched
   * @param formGroup - The form group to touch
   */
private markFormGroupTouched(formGroup: FormGroup) {
  (<any>Object).values(formGroup.controls).forEach(control => {
    control.markAsTouched();
    control.updateValueAndValidity(); // <-- I added this line
    if (control.controls) {
      this.markFormGroupTouched(control);
    }
  });
}
```

and the line as this:
```js
(this.pickerInputTouched || !!this.ngControl.touched) &&
```

Otherwise, without the `ngControl.touched`, even if the developer set the formControl as invalid and touched, he won't be able to display the error message programatically since he cannot change the value of `pickerInputTouched`...

christophercr: I've found a simpler solution... which just requires the developer to mark the formControl as `dirty` or `touched`.

I'll update the PR 😉

christophercr: Ok, this is the solution I implemented:

```typescript
/**
 * Part of {@link MatFormFieldControl} API
 * @ignore
 * @internal
 */
public get errorState(): boolean {
	// the control can be in an error state as long as one of these conditions is met:
	// 1) the user has interacted with it
	// 2) the control is programmatically marked as 'touched' or 'dirty'
	const newErrorState =
		this.ngControl !== null &&
		this.ngControl.control !== null &&
		(this.pickerInputTouched || !!this.ngControl.touched || !!this.ngControl.dirty) &&
		(!!this.ngControl.errors || !!this.pickerInput.validate(this.ngControl.control));

	// IMPORTANT: emit an state change when the errorState changes
	// This is needed to force the MatFormFieldControl to refresh and render the MatError's
	if (this._errorState !== newErrorState) {
		this._errorState = newErrorState;
		this.stateChanges.next();
	}
		return this._errorState;
}

/**
 * The current error state
 * @ignore
 * @internal
 */
public _errorState = false;
```

So indeed, as you mentioned we need to check for `ngControl.touched` and `ngControl.dirty`... but the most important part to make the errors appear is to emit on the `stateChanges`.

christophercr:
PR updated with the fixes for your remarks.

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
SuperITMan: That seems to be a good solution :smiley:

packages/stark-ui/src/modules/date-picker/components/date-picker.component.ts
```diff
@@ -229,40 +247,69 @@ export class StarkDatePickerComponent extends AbstractStarkUiComponent
...
-		return (
+		// the control can be in an error state as long as one of these conditions is met:
+		// 1) the user has interacted with it
+		// 2) the control is programmatically marked as 'touched' or 'dirty'
+		const newErrorState =
 			this.ngControl !== null &&
 			this.ngControl.control !== null &&
-			(!!this.ngControl.errors || !!this.pickerInput.validate(this.ngControl.control))
-		);
+			(this.pickerInputTouched || !!this.ngControl.touched || !!this.ngControl.dirty) &&
```
SuperITMan: Small typo: emit **a** state change

christophercr: oops... I'll fix it right away!

Issue #1260, opened 2019-04-23
ui: datePicker - When using starkDateMask, the date is broken when edit it


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When click on the date to edit it: 
  * April 23, 2019 becomes 23/01/2019
  * When edit 23/01/2019 to 23/04/2019, it becomes April 23, 0019

![2019-04-23_10-21-56](https://user-images.githubusercontent.com/10333565/56566007-012fba80-65b2-11e9-802e-b52f623e6724.gif)

## Expected behavior

The date should not be broken by the date mask.

## Minimal reproduction of the problem with instructions

Go in the demo app, in demo-ui > components > date-picker and reproduce the same steps than the one if the video above.

## Environment
<pre><code>
Stark version: 10.0.0-beta.7
</code></pre>

christophercr:
After investigating the issue, the problem is due to 2 things:

1. wrong behavior of the mask when deleting characters. The mask moves the rest of the characters backwards/forwards which causes the date picker to get weird date values.
2. the translation initialization logic is not correct. The locale set in Moment is `en-gb` but it should be `en` (US English). This is because the last imported locale sets the current locale (in the Showcase/Starter it was `en-gb`).

Apart from this, we should improve the logic when the application language changes because right now the Moment locale is not changed, which means that the date pickers don't show the date in the current language :(

christophercr:
I've created this PR (#1271) for the different fixes

christophercr:
The issue about the wrong behavior when deleting characters will be partially fixed in #1271 so we can release soon.

However, this should be completely fixed later on. This will be tracked in #1277

PR #1271, opened 2019-04-30
Bugfix/date picker mask weird behavior + inconsistent date picker locale in Showcase and Starter

christophercr:
ISSUES CLOSED: #1260 

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1260

## What is the new behavior?
- now when deleting a character, the mask keeps the position as empty instead of moving the characters backwards so the user can safely type in the new character. This prevents the date picker from getting weird date values. The `keepCharPositions` option is now enabled in the timestamp directive (see https://github.com/text-mask/text-mask/blob/master/componentDocumentation.md#keepcharpositions)
- the max and min range values for the different date parts of a timestamp are now correct.
- expanded the unit tests of the `createTimestampPipe` to cover all possible use cases with date time strings.
- enhanced the translation initialization to also initialize the Moment locale to keep it in sync with the application language.
- implemented some logic to change the Moment locale and the Material Date Adapter when the app language changes.

**IMPORTANT: Although the locale of the Material Date Adapter is now changed when the app language changes, the date pickers don't reflect the new locale. This is a known bug in Angular Material (see https://github.com/angular/components/issues/15419) and should be fixed soon.**

**So for the moment, the logic has been put in the `SharedModule` instead to make it work**

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-ui/src/modules/input-mask-directives/directives/timestamp-mask.directive.ts
+++ b/packages/stark-ui/src/modules/input-mask-directives/directives/timestamp-mask.directive.ts
@@ -122,7 +122,8 @@ export class StarkTimestampMaskDirective extends MaskedInputDirective implements
 			return {
 				pipe: <any>createTimestampPipe(timestampMaskConfig.format),
 				mask: this.convertFormatIntoMask(timestampMaskConfig.format),
-				placeholderChar: "_"
+				placeholderChar: "_",
+				keepCharPositions: true // to avoid weird date values when deleting characters (see https://github.com/NationalBankBelgium/stark/issues/1260)
 			};
 		}
 	}
--- a/showcase/src/app/app.module.ts
+++ b/showcase/src/app/app.module.ts
@@ -263,7 +263,7 @@ export class AppModule {
 
 		this.settingsService.initializeSettings();
 
-		sessionService
+		this.sessionService
 			.getCurrentUser()
 			.pipe(filter((user?: StarkUser): user is StarkUser => typeof user !== "undefined"))
 			.subscribe((user: StarkUser) => {
--- a/starter/src/app/app.module.ts
+++ b/starter/src/app/app.module.ts
@@ -218,7 +218,7 @@ export class AppModule {
 
 		this.settingsService.initializeSettings();
 
-		sessionService
+		this.sessionService
 			.getCurrentUser()
 			.pipe(filter((user?: StarkUser): user is StarkUser => typeof user !== "undefined"))
 			.subscribe((user: StarkUser) => {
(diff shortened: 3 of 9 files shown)


christophercr:
Could you have a look on this PR?

showcase/src/app/app.module.ts
```diff
@@ -263,12 +266,33 @@ export class AppModule {
...
 			.pipe(filter((user?: StarkUser): user is StarkUser => typeof user !== "undefined"))
 			.subscribe((user: StarkUser) => {
 				const devAuthenticationHeaders: Map<string, string> = getAuthenticationHeaders(user);
 				this.sessionService.setDevAuthenticationHeaders(devAuthenticationHeaders);
 			});
+
+		/**
+		 * When the application language changes, the Moment locale and the Material Moment Adapter (internally used by Stark Date Pickers) should be changed as well
+		 */
+		this.sessionService
```
SuperITMan: It's unfortunately not working when it's part of the **App** module. Could you please move this logic in the **Shared** module to make it work ? 😊

starter/src/app/app.module.ts
```diff
@@ -218,12 +221,33 @@ export class AppModule {
...
 			.pipe(filter((user?: StarkUser): user is StarkUser => typeof user !== "undefined"))
 			.subscribe((user: StarkUser) => {
 				const devAuthenticationHeaders: Map<string, string> = getAuthenticationHeaders(user);
 				this.sessionService.setDevAuthenticationHeaders(devAuthenticationHeaders);
 			});
+
+		/**
+		 * When the application language changes, the Moment locale and the Material Moment Adapter (internally used by Stark Date Pickers) should be changed as well
+		 */
+		this.sessionService
```
SuperITMan: Same here about moving this in the **Shared** module

packages/stark-ui/src/modules/input-mask-directives/directives/timestamp-mask.directive.ts
```diff
@@ -122,7 +122,8 @@ export class StarkTimestampMaskDirective extends MaskedInputDirective implements
 			return {
 				pipe: <any>createTimestampPipe(timestampMaskConfig.format),
 				mask: this.convertFormatIntoMask(timestampMaskConfig.format),
-				placeholderChar: "_"
+				placeholderChar: "_",
+				keepCharPositions: true // to avoid weird date values when deleting characters (see https://github.com/NationalBankBelgium/stark/issues/1260)
```
SuperITMan: It's not really working fine... If we have 13/05/2019 and we want to change the month for December, we cannot do it.
We can obtain 13/0_/2019 but not 13/__/2019... Maybe it's already better than nothing 😊

christophercr: As discussed, I've partially fixed this issue. There are still some edge cases that are not covered but I think they are not urgent for now so we can release soon.

So I've created #1277 to track the issue for those edge cases that we should fix later on 😉

christophercr:
@SuperITMan PR updated with the remarks. Can you review them?

Issue #1283, opened 2019-05-17
ui: table - Add option to show row index


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

No row index available.

## Expected behavior

In the previous implementation of Stark, there was an option `showRowIndex` to display the row index automatically. It should be there again.

## What is the motivation / use case for changing the behavior?

Provide options available before.

PR #1286, opened 2019-05-19
fix(stark-ui): table - add row index option

SuperITMan:
ISSUES CLOSED: #1283

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1283 

## What is the new behavior?
Provide a row index.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/table/components/table.component.spec.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.spec.ts
@@ -22,6 +22,7 @@ import { StarkTableComponent } from "./table.component";
 import { StarkTableColumnComponent } from "./column.component";
 import { StarkPaginationComponent } from "../../pagination/components";
 import { StarkTableColumnFilter, StarkTableColumnProperties, StarkTableFilter, StarkTableRowActions } from "../entities";
+import find from "lodash-es/find";
 import Spy = jasmine.Spy;
 import createSpy = jasmine.createSpy;
 
@@ -36,9 +37,10 @@ import createSpy = jasmine.createSpy;
 			[multiSort]="multiSort"
 			[rowsSelectable]="rowsSelectable"
 			[multiSelect]="multiSelect"
+			[showRowsCounter]="showRowsCounter"
+			[showRowIndex]="showRowIndex"
 			[orderProperties]="orderProperties"
 			[tableRowActions]="tableRowActions"
-			[showRowsCounter]="showRowsCounter"
 			[rowClassNameFn]="rowClassNameFn"
 			(rowClicked)="rowClickHandler($event)"
 		>
@@ -56,6 +58,7 @@ class TestHostComponent {
 	public multiSelect?: string;
 	public multiSort?: string;
 	public showRowsCounter?: boolean;
+	public showRowIndex?: boolean;
 	public tableRowActions?: StarkTableRowActions;
 	public tableFilter?: StarkTableFilter;
 	public orderProperties?: string[];
@@ -154,11 +157,14 @@ describe("TableComponent", () => {
 			expect(component.multiSort).toBe(hostComponent.multiSort);
 			expect(component.orderProperties).toBe(hostComponent.orderProperties);
 			expect(component.showRowsCounter).toBe(false);
+			expect(component.showRowIndex).toBe(false);
 			expect(component.tableRowActions).toBe(<any>hostComponent.tableRowActions);
 		});
 	});
 
 	describe("on change", () => {
+		const tableThSelector = ".stark-table thead tr th";
+
 		it("should make a copy of 'data' in 'dataSource' when 'data' changes to keep 'data' immutable", () => {
 			const dummyData = [{ name: "test-data" }];
 			hostComponent.dummyData = dummyData;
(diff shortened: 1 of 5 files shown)


packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -224,6 +225,34 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	@Input()
 	public rowClassNameFn?: (row: object, index: number) => string;
 
+	/**
+	 * Determine if the row index must be present or not.
+	 * Default: true
```
cnomes: Shouldn't this be false?

christophercr: In fact, what's wrong in the source code... it should be initialized as `true` :p

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -224,6 +225,34 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
+			const i: number = this.displayedColumns.indexOf("rowIndex");
+			this.displayedColumns.splice(i);
+		}
+	}
+
+	/**
+	 * @ignore
+	 * @internal
+	 */
+	public _showRowIndex = false;
```
christophercr: This should be initialized as `true` (as we have it in Stark 9)

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -236,6 +239,21 @@ describe("TableComponent", () => {
 			expect(component.displayedColumns.unshift).not.toHaveBeenCalled();
 		});
 
+		it("should assign right value to _showRowIndex when showRowIndex changes and adapt displayedColumns", () => {
+			spyOn(component.displayedColumns, "unshift");
+			hostComponent.showRowIndex = true;
+			hostFixture.detectChanges();
+			expect(component._showRowIndex).toBe(true);
+			expect(component.displayedColumns.unshift).toHaveBeenCalledTimes(1);
+			expect(component.displayedColumns.unshift).toHaveBeenCalledWith("rowIndex");
```
christophercr: I think the test should focus on the functionality rather than on the implementation. In this case we should not test that `unshift` was called because this is an implementation detail. The test should focus on the removal and the changes on the `displayedColumns` 😉

packages/stark-ui/src/modules/table/components/table.component.ts
SuperITMan: Previously, it was `true` by default. Do we keep the same behavior so ? 😊

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -939,6 +968,16 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 		}
 	}
 
+	/**
+	 * Get the row index, based on its position in dataSource.data
+	 * @param row - Row to get index
+	 */
+	public getRowIndex(row: any): number | void {
```
christophercr: I think we should return `undefined` instead of `void` so it is clear that no index was found. Normally `void` is used when the flow of the program/function does not return anything... but in this case we should always return something...

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -971,6 +989,19 @@ describe("TableComponent", () => {
...
+	describe("getRowIndex", () => {
+		it("should return the right index for every row", () => {
+			for (let i = 0; i < component.dataSource.data.length; i++) {
+				expect(component.getRowIndex(component.dataSource.data[i])).toBe(i+1);
+			}
+		});
+		
+		it("should return 'void' if dataSource is not initialized yet", () => {
+			component.dataSource = <any>undefined;
+			expect(component.getRowIndex({})).toBe();
```
christophercr: Small change: can you add some dummy data to the `{}` so it is clear that is just a row object?

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -971,6 +989,19 @@ describe("TableComponent", () => {
 			expect(component.getColumnSortingPriority("c")).toBeUndefined();
 		});
 	});
+	
+	describe("getRowIndex", () => {
+		it("should return the right index for every row", () => {
+			for (let i = 0; i < component.dataSource.data.length; i++) {
+				expect(component.getRowIndex(component.dataSource.data[i])).toBe(i+1);
```
christophercr: Can you run prettier on this file? It seems it is not correctly formatted.

packages/stark-ui/src/modules/table/components/table.component.ts
christophercr: Yep, I think we should keep the same behavior as in Stark 9...

cnomes: To me it makes more sense that it's `false` by default.
And it also seems to work like this at the moment

packages/stark-ui/src/modules/table/components/table.component.spec.ts
SuperITMan: Actually, I applied the same logic than the one we applied for `multiSelect` when we check if the checkbox is displayed or not by checking the `displayedColumns`. 
I'll try to improve this 😊

packages/stark-ui/src/modules/table/components/table.component.ts
SuperITMan: Currently, it's false by default (I didn't want to force everyone to have an index by default).
But, in the previous implementation of Stark (v9), it was enabled by default... 
I don't feel comfortable to enable this feature by default actually 😕

packages/stark-ui/src/modules/table/components/table.component.spec.ts
christophercr: Yep, that's fine... I'm just saying that we should avoid testing the internal functions we call such as `unshift` cause maybe we might change that later on and our test will fail even though the functionality is still correct :s

packages/stark-ui/src/modules/table/components/table.component.ts
christophercr: Ok, so let's keep it as disabled by default in Stark 10 😉

SuperITMan:
@cnomes @christophercr I upaded my PR based on your remarks. Could you please have a look ? :blush:

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -220,20 +226,63 @@ describe("TableComponent", () => {
...
-			expect(component.isMultiSelectEnabled).toBe(false);
 			expect(component.displayedColumns.unshift).not.toHaveBeenCalled();
+			rowThElements = <NodeListOf<HTMLElement>>hostFixture.nativeElement.querySelectorAll(tableThSelector);
+			expect(rowThElements.length).toBeGreaterThanOrEqual(0);
+			selectThElement = find(rowThElements, (thElement: HTMLElement) => thElement.className.indexOf("cdk-column-select") > -1);
+			expect(selectThElement).toBeUndefined();
+		});
+
+		it("should assign right value to _showRowIndex when showRowIndex changes and adapt displayedColumns", () => {
+			spyOn(component.displayedColumns, "unshift").and.callThrough();
```
christophercr: Can you remove this line from this test? As I mentioned before, we should avoid testing these kind of "internals"

SuperITMan:
@christophercr @cnomes PR udpated based on your remarks 😊

Issue #1303, opened 2019-05-29
ui: table - add a visual indicator when the global filter is enabled


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When the global filter is enabled in the table component, there is no indicator whatsoever about such filtering, which might be confusing for the end user when the data in the table is refreshed and he might expect to see all the rows but he will only see the filtered ones.

## Expected behavior

A visual indicator should be in place so the user knows that he as the global filter enabled. **This is already the case for column filters: the filter icon is gray or black when it is disabled or enabled.**

Something similar can be applied to the global filter button, but in that case it could be blue when enabled and black when disabled. Like in the Network tab in Chrome Dev tools:

![filter indicator chrome devtools](https://user-images.githubusercontent.com/5698318/58569692-e5fe3d80-8236-11e9-907e-c2971180306f.PNG)

## Minimal reproduction of the problem with instructions

1. Go to https://stark.nbb.be/showcase/latest/demo-ui/table and enable the global filter in any of the tables.
2. Click on any action, paginate the data, etc. There is no clear indication that the global filter is enabled.

## What is the motivation / use case for changing the behavior?

Better UX.

## Environment

<pre><code>
Stark version: 10.0.0-beta.7

</code></pre>

christophercr:
As discussed with @SuperITMan, to keep consistency, we will apply the same color change as we do with the filter icon in the columns (black when the filter is applied and gray when it is not).

PR #1310, opened 2019-06-03
feat(stark-ui): change color of global filter icon to indicate when the global filter is enabled

christophercr:
ISSUES CLOSED: #1303

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1303 

## What is the new behavior?
The icon of the global filter now changes color whenever the filter is enabled so the user knows that the filtered is being applied.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-ui/src/modules/table/components/_table-theme.scss
+++ b/packages/stark-ui/src/modules/table/components/_table-theme.scss
@@ -1,7 +1,7 @@
 /* ============================================================================== */
-/*  S t a r k   A p p   T a b l e                                                  */
+/*  S t a r k   T a b l e                                                         */
 /* ============================================================================== */
-/* stark-ui: src/modules/app-menu/components/_table-theme.scss */
+/* stark-ui: src/modules/table/components/_table-theme.scss */
 
 .stark-table {
   tr {
--- a/packages/stark-ui/src/modules/table/components/_table.component.scss
+++ b/packages/stark-ui/src/modules/table/components/_table.component.scss
@@ -1,7 +1,7 @@
 /* ============================================================================== */
 /*  S t a r k   T a b l e                                                         */
 /* ============================================================================== */
-/* stark-ui: src/modules/table/components/table/_table.component.scss */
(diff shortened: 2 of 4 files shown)


SuperITMan:
What do you think about following the same approach that we follow for the columns ?
Currently, your change proposes this:
image

But what if we follow the same approach than the one we have for the column ?
```css
.stark-table .header .actions {
  .button-global-filter {
    opacity: 0.3;
    &.filter-enabled {
      opacity: 1;
    }
  }
}
```

image

What do you think @cnomes @christophercr @nicanac ? 😊

christophercr:
Indeed, it is more consistent that way. I'll change it right now! 😉

christophercr:
@SuperITMan @cnomes @nicanac I've updated the PR with the change of color to black/gray. Can you have a look?

packages/stark-ui/src/modules/table/components/_table-theme.scss
```diff
@@ -1,7 +1,7 @@
 /* ============================================================================== */
-/*  S t a r k   A p p   T a b l e                                                  */
+/*  S t a r k   A p p   T a b l e                                                 */
```
SuperITMan: Mmm... I think it should be 
```txt
S t a r k   T a b l e
```
😛

packages/stark-ui/src/modules/table/components/_table.component.scss
```diff
@@ -19,6 +19,13 @@
       display: flex;
       justify-content: flex-end;
 
+      .button-global-filter {
+        opacity: 0.2;
```
SuperITMan: Why did you set `opacity: 0.2` and not `0.3` as is this the case for the columns ?
```css
.button-filter {
  opacity: 0.3;
}
```

christophercr: Yeah, I know, but as I explained to you, the opacity should be lighter because the base color in the header is darker than in the columns 😉

Issue #1364, opened 2019-07-26
ui: dropdown : no take into account option "emitEvent: false"


nicanac:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior
Option "emitEvent: false" is not taken into account by Stark Dropdown.

When set the Stark Dropdown component disable /enable with `{ emitEvent: false }`, the valueChanges is triggered...

 
```typescript
this.formGroup.controls["hierarchicalFlag"].disable({ emitEvent: false });
 
this.hierarchicalGroup.controls["codeHierarchical"].valueChanges
   .pipe(
      takeUntil(this.destroyed$),
      distinctUntilChanged()
   )
   .subscribe((value: string) => { // an event is emmited even if emitEvent is set false
      this.onCodeHierarchicalChange(value);
   });
```

## Expected behavior
Not triggering the valueChanges if `emitEvent: false` is set.

## What is the motivation / use case for changing the behavior?

Stable components.

## Environment

<pre><code>
Stark version: 10.0.0_beta.8

</code></pre>

PR #1365, opened 2019-07-26
fix(stark-ui): ui dropdown: not taking into account option - emitEvent

nicanac:
ISSUES CLOSED: #1364

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Change the dropdown setDisabledState function to take into account the emitEvent.

Issue Number: #1364 

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
+++ b/packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
@@ -559,6 +559,48 @@ describe("DropdownComponent", () => {
 					hostFixture.detectChanges();
 					expect(selectElement.outerHTML).toMatch(/ng-reflect-disabled="false"/);
 				});
+
+				it("shouldn't trigger a 'valueChange' event when the formControl is disabled or enabled with emitEvent set to false", () => {
+					const mockObserver: SpyObj<Observer<any>> = createSpyObj<Observer<any>>("observerSpy", ["next", "error", "complete"]);
+
+					hostComponent.formControl.valueChanges.subscribe(mockObserver);
+
+					hostComponent.formControl.disable({ emitEvent: false });
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(true);
+
+					hostComponent.formControl.enable({ emitEvent: false });
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(false);
+					expect(mockObserver.next).not.toHaveBeenCalled(); // because the 'emitEvent' is false
+					expect(mockObserver.error).not.toHaveBeenCalled();
+					expect(mockObserver.complete).not.toHaveBeenCalled();
+				});
+
+				it("should trigger a 'valueChange' event when the formControl is disabled or enabled with emitEvent set to true", () => {
+					const mockObserver: SpyObj<Observer<any>> = createSpyObj<Observer<any>>("observerSpy", ["next", "error", "complete"]);
+
+					hostComponent.formControl.valueChanges.subscribe(mockObserver);
+
+					hostComponent.formControl.disable(); // 'emitEvent' true by default
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(true);
+					expect(mockObserver.next).toHaveBeenCalledTimes(1);
+					expect(mockObserver.error).not.toHaveBeenCalled();
+					expect(mockObserver.complete).not.toHaveBeenCalled();
+					mockObserver.next.calls.reset();
+
+					hostComponent.formControl.enable(); // 'emitEvent' true by default
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(false);
+					expect(mockObserver.next).toHaveBeenCalledTimes(1);
+					expect(mockObserver.error).not.toHaveBeenCalled();
+					expect(mockObserver.complete).not.toHaveBeenCalled();
+				});
 			});
 
 			describe("formControl.value", () => {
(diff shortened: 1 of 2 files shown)


packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -561,6 +561,47 @@ describe("DropdownComponent", () => {
 				});
 			});
 
+			describe("DropDown properties", () => {
```
SuperITMan: Could you please change the description for something more related to what you will test ? 😊

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -561,6 +561,47 @@ describe("DropdownComponent", () => {
...
+				let mockObserver: SpyObj<Observer<any>>;
+
+				beforeEach(() => {
+					mockObserver = createSpyObj<Observer<any>>("observerSpy", ["next", "error", "complete"]);
+					component = hostComponent.dropdownComponent;
```
SuperITMan: This line is not needed since it's already done in the `beforeEach` at line 500

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -561,6 +561,47 @@ describe("DropdownComponent", () => {
...
+				});
+
+				it("the MatOption should be disabled when the form control is disabled AND it should trigger a 'valueChange' event ONLY IF the 'emitEvent' option is enabled", () => {
```
SuperITMan: Could you please split this test in 2 parts (one with `emitEvent: false`, the other with `emitEvent: true` ?
Also, could you rephrase the `it` ?  
What you want to test here is that when you disable the `formControl` with(out) `emitEvent`, it reacts as expected. 
You don't care at this moment if the MatOption is disabled or not. Actually this is checked already at line 552 😊

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -561,6 +561,46 @@ describe("DropdownComponent", () => {
...
+				});
+
+				it("When the formControl is disabled or enabled with emitEvent with emitEvent set to true the dropdown shouldn't trigger a 'valueChange' event", () => {
+					hostComponent.formControl.valueChanges.subscribe(mockObserver);
+
+					hostComponent.formControl.disable(); // 'emitEvent' true by default
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(true);
+					expect(mockObserver.next).toHaveBeenCalledTimes(1);
```
SuperITMan: Could you please add those 2 checks ?
```typescript
expect(mockObserver.error).not.toHaveBeenCalled();
expect(mockObserver.complete).not.toHaveBeenCalled();
```

Otherwise, if they are called after (line 599), you cannot be sure they were not called before...

nicanac: added

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -561,6 +561,46 @@ describe("DropdownComponent", () => {
 				});
 			});
 
+			describe("Changing emitEvent properties ", () => {
```
SuperITMan: This `describe` is not really relevant since you are not "changing emitEvent properties". Actually, you are disabling/enabling the dropdown with or without the `emitEvent; false`.
Maybe you could move your 2 `it` into `describe("formControl.disabled", ()`. What do you think @nicanac ?

In that case, I would advise to you to declare the mockObserver in each `it` since it's not necessary for 
```
it("should set the right value to the mat-select when 'formControl.disabled' changes", () => {
```

nicanac: yes, why not, seems legit for me.

nicanac: moved to the `describe("formControl.disabled", ().`

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -559,6 +559,48 @@ describe("DropdownComponent", () => {
 					hostFixture.detectChanges();
 					expect(selectElement.outerHTML).toMatch(/ng-reflect-disabled="false"/);
 				});
+
+				it("Shouldn't trigger a 'valueChange' event when the formControl is disabled or enabled with emitEvent set to false", () => {
```
SuperITMan: could you please change this in lowercase as it is the case in the other tests ? 😊 
```
it("shouldn't trigger a 'valueChange' event when the formControl is di ...
```

packages/stark-ui/src/modules/dropdown/components/dropdown.component.spec.ts
```diff
@@ -559,6 +559,48 @@ describe("DropdownComponent", () => {
...
+					hostComponent.formControl.enable({ emitEvent: false });
+					hostFixture.detectChanges();
+
+					expect(component.disabled).toBe(false);
+					expect(mockObserver.next).not.toHaveBeenCalled(); // because the 'emitEvent' is false
+					expect(mockObserver.error).not.toHaveBeenCalled();
+					expect(mockObserver.complete).not.toHaveBeenCalled();
+				});
+
+				it("Shouldn't trigger a 'valueChange' event when the formControl is disabled or enabled with emitEvent set to true", () => {
```
SuperITMan: I think that it "should" trigger in this case :wink:

christophercr: "should trigger ..."

Issue #1366, opened 2019-07-30
ui: table - selection is not reset when data changes


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When data changes in table, the selection is not reset. 

## Expected behavior

When the data changes, the previous selection becomes invalid and should be reset.

## What is the motivation / use case for changing the behavior?

Fix bug

## Environment

<pre><code>
Stark version: 10.0.0-beta.8

</code></pre>

PR #1367, opened 2019-07-30
fix(stark-ui): table - when data changes, selection is now reset

SuperITMan:
ISSUES CLOSED: #1366

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Selection is not reset when data changes in table component

Issue Number: #1366 

## What is the new behavior?
Selection is reset when data changes.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/table/components/table.component.spec.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.spec.ts
@@ -209,6 +209,16 @@ describe("TableComponent", () => {
 			expect(component.dataSource.data).toEqual([{ name: "test-data-2" }]);
 		});
 
+		it("should reset 'selection' when 'data' changes", () => {
+			component.selection.toggle({ name: "selected-data-1" });
+			expect(component.selection.selected).toEqual([{ name: "selected-data-1" }]);
+
+			hostComponent.dummyData = [{ name: "data-1" }];
+			hostFixture.detectChanges();
+
+			expect(component.selection.selected).toEqual([]);
+		});
+
 		it("should assign right value to isFixedHeaderEnabled when fixedHeader changes", () => {
 			hostComponent.fixedHeader = "true";
 			hostFixture.detectChanges();
--- a/packages/stark-ui/src/modules/table/components/table.component.spec.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.spec.ts
@@ -40,6 +40,7 @@ import createSpy = jasmine.createSpy;
 			[showRowsCounter]="showRowsCounter"
 			[showRowIndex]="showRowIndex"
 			[orderProperties]="orderProperties"
+			[selectedRows]="selectedRows"
 			[tableRowActions]="tableRowActions"
 			[rowClassNameFn]="rowClassNameFn"
 			(rowClicked)="rowClickHandler($event)"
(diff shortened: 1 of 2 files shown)


SuperITMan:
@carlo-nomes It's true... 
The annoying issue here is that the developer has no way to reset the selection when he wants.
Or maybe we could add the input "selection". Like this, the dev can control the selection if he wants and we can still manage it ourselves in Stark Table. 

What do you think ? 😊

carlo-nomes:
@SuperITMan that is indeed a possibility, we could make it a two way binding. so you can add something like this: `<stark-table [(selection)]="selection"></stark-table>`. But it might be more difficult than I'm thinking, because of object references. 🤔 

My first thought was actually simpler (and less complete). We could make the `resetSelection` function public. So that developers can access it via a ViewChild.

christophercr:
Well, in my opinion just having a `selection` input would be enough cause that could also be used to reset the selection, right? Although as @carlo-nomes suggested, I'm not sure if the selection model can deal with old object references...

carlo-nomes:
I suppose it will be an issue when people deep clone or recreate their data. Maybe we can provide a way to let them choose a "selection identifier"? 🤔 

```html 
<stark-table 
  [data]="data"
  [(selection)]="selection" 
  [selectionIdentifier]="selectionIdentifier">
</stark-table>
```
```JavaScript
//...
public class Component {
  data = [
    { id: "1", desc: "something" },
    { id: "2", desc: "something" },
    { id: "3", desc: "something" },
  ]

  selection = [this.data[0]]

  function selectionIdentifier(item){
    return item.id;
  }
}
```

SuperITMan:
I made a try by adding an input "selectedRows" as it was the case in the previous Stark.

@carlo-nomes Actually Angular Material does not care about the ID of a row. It seems it compares the entire row to know if it is selected or not...

@christophercr @nicanac @carlo-nomes Could you please have a look on this new implementation and tell me what you think about it ? 😊 

I'll remove the first commit after. Like this you can easily compare what I changed.

SuperITMan:
My bad, it's not working at all >.<
I'll update the PR when it is fixed 😊

SuperITMan:
@christophercr @carlo-nomes @nicanac 
I understood how it is working. Let's take this example to show how it can work or not.
```typescript
const data = [
  { id: 1, name: "name-1", value: "value-1" },
  { id: 2, name: "name-2", value: "value-2" },
  { id: 3, name: "name-3", value: "value-3" },
  { id: 4, name: "name-4", value: "value-4" }
]

const selectedRows = [];

selectRow1() {
  // the row is not selected in the table
  this.selectedRows = [{ id: 1, name: "name-1", value: "value-1" }];
}

selectRow2() {
  // the row is correctly selected in the table
  this.selectedRows = [data[1]];
}
```

```html
<stark-table
  [data]="data"
  multiSelect
  ...
  [selectedRows]="selectedRows">
</stark-table>
```

Do you think we can keep this implementation or should we improve it ?

carlo-nomes:
Could we make it a 2-way binding? Because it might be confusing that the list in their own class is not up to date with the actual selection. 🤔

SuperITMan:
I close this in favor of my new PR #1377 which is based on `selection` object from angular/material 😊

Issue #1338, opened 2019-06-18
showcase: header not correctly displayed on Safari


christophercr:
## I'm submitting a...

<pre><code>
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When opening the Showcase (https://stark.nbb.be/showcase/latest/home) in Safari, the header is not displayed correctly, just a big blank space instead.

<img width="1414" alt="AppHeader issue on Safari" src="https://user-images.githubusercontent.com/5698318/59697487-2f1a2f80-91ee-11e9-823c-08f51466b4b7.png">

## Expected behavior

The Showcase should be displayed correctly in all evergreen browsers.

## What is the motivation / use case for changing the behavior?

Cross-browser compatibility

## Environment

<pre><code>
Angular version: 7.2.15
Stark version: 10.0.0-beta.8

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [X] Safari (desktop) version 12.1.1
</code></pre>

SuperITMan:
Re-open this since it's still happening on iOS (iPad & iPhone)

iPhone 8:
![image](https://user-images.githubusercontent.com/10333565/73457077-5edc0900-4373-11ea-8309-50c4ffac00ba.png)

iPad Pro 12' 2018:
![image](https://user-images.githubusercontent.com/10333565/73457177-8e8b1100-4373-11ea-8ede-f4ff2bf97407.png)

PR #1381, opened 2019-08-13
fix(stark-ui): fix side navigation rendering behind header

carlo-nomes:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Because of a bug previously introduced in a previous commit the side navigations (left / right) were rendering behind the header, causing issues on mobile or whenever the side navigation has full page height.

## What is the new behavior?
This commit fixes the initial Safari bug (#1338) without breaking the side navigation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/stark-ui/assets/styles/_base.scss
+++ b/packages/stark-ui/assets/styles/_base.scss
@@ -17,15 +17,28 @@ body {
     margin-bottom: 64px;
   }
 
-  main.stark-main-container {
-    flex: 1 0 auto;
-    align-self: center;
+  .mat-drawer {
+    z-index: 120; /* z-index of header is 110. z-index of mat-drawer should be bigger */
+  }
+  
+  .mat-drawer-content {
+    z-index: initial; // fix safari not showing fixed header
 
-    width: 100%;
-    max-width: $stark-max-content-width;
+    display: flex;
+    flex-direction: column;
 
-    box-sizing: border-box;
-    padding: 64px 15px;
+    overflow: auto;
+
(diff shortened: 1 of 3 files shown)


nicanac:
Works well on my local.
But I noticed an issue with the travis' build.
Maybe a simple commit amend is enough.

christophercr:
Can you guys double check in Firefox? In my case (on Mac) it seems is not possible to scroll anymore on the screens that are bigger than the viewport, for example the `demo-ui/input-mask-directives`page

packages/stark-ui/assets/styles/_base.scss
```diff
@@ -17,15 +17,29 @@ body {
     margin-bottom: 64px;
   }
 
-  main.stark-main-container {
-    flex: 1 0 auto;
-    align-self: center;
+  .mat-drawer-content {
+    overflow: visible;
```
carlo-nomes: @christophercr you are right, this line makes Firefox not respect the `height: 100%` of `.mat-drawer-content` (making it impossible to scroll in the `.stark-scroll-container`). The problem is... if I put set this to `overflow: hidden` Safari does not show the `<header>`. Do you or @nicanac or @SuperITMan  have any suggestion on the best way to resolve this? 🤔

carlo-nomes:
Okay, I think I found / fixed the source of the problem. That was a difficult one. 😅Could you review again?

Issue #1341, opened 2019-06-21
ui: app-data - in 'dropdown' mode the 'detail' pane overlaps the 'summary' pane if the summary is too big


christophercr:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

In the AppData in 'dropdown' mode, when the 'summary' pane is too big, it is overlapped by the 'detail' pane so the summary content is not fully visible:

<img width="285" alt="AppData summary" src="https://user-images.githubusercontent.com/5698318/59934485-ef508380-944b-11e9-831e-5bdd1da82e34.png">
<img width="285" alt="AppData summary overlapped by detail" src="https://user-images.githubusercontent.com/5698318/59934488-f24b7400-944b-11e9-983a-4dab5fe5684c.png">

## Expected behavior

The 'detail' pane should not overlap the 'summary' pane regardless of its height.

## Minimal reproduction of the problem with instructions

1. Open the AppData demo page in the Showcase: https://stark.nbb.be/showcase/latest/demo-ui/app-data
2. Change the language of the app to French or Dutch
3. In the first example, click on the dropdown to expand the AppData component to show the 'detail' pane.
4. The detail pane overlaps with the summary pane because the summary is too big due to the French/Dutch text.

## What is the motivation / use case for changing the behavior?

Correct display of the AppData component regardsless of its content.

## Environment

<pre><code>
Stark version: 10.0.0-beta.8

</code></pre>

PR #1383, opened 2019-08-13
fix(stark-ui): ui: app-data - in 'dropdown' mode the 'detail' pane overlaps the 'summary'

nicanac:
In the AppData in 'dropdown' mode, when the 'summary' pane is too big, it is overlapped by the
'detail' pane so the summary content is not fully visible

ISSUES CLOSED: #1341

## PR Checklist
The 'summary' pane isn't overlapped anymore.

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1341 

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/app-data/components/_app-data.component.scss
+++ b/packages/stark-ui/src/modules/app-data/components/_app-data.component.scss
@@ -46,7 +46,7 @@
 
   & .stark-app-data-detail {
     position: absolute;
-    top: 55px;
+    top: 100%;
     width: 250px;
     right: 0;
     left: 0;


nicanac:
> I don't think that we should solve it by setting a fixed number to the `top`or `bottom` properties, otherwise, the issue will come back in case more info is added to the summary and gets bigger.
> 
> I suggest to solve it by setting the `top`or `bottom` dynamically according to the height of the summary pane. What do you think @SuperITMan @carlo-nomes?

What browser are you using ? Because  it solve the issue. Now, with the bottom positioning the 'detail' pane is always positioned regarding the bottom of the parent container. So, whatever the size, it'll never overlap the summary.

christophercr:
Oops, my bad, I didn't see the `%` sign... sorry

Then indeed, it solves the issue 👍

Issue #1214, opened 2019-03-21
showcase: add anchors on every example in documentation.


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

It's difficult to share an example from the showcase.
If you want to share a link to show `Table with fixed row actions`, you can only send https://stark.nbb.be/showcase/latest/demo-ui/table to your contact.

## Expected behavior

It would be nice if every example had an anchor, like this we could share:
https://stark.nbb.be/showcase/latest/demo-ui/table#table-fixed-row-actions 

## What is the motivation / use case for changing the behavior?

Improve the showcase

PR #1388, opened 2019-08-20
feat(stark-demo): add anchors to examples

carlo-nomes:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1214 

## What is the new behavior?
Added anchors to all examples

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/stark-core/src/modules/routing/services/routing.service.ts
+++ b/packages/stark-core/src/modules/routing/services/routing.service.ts
@@ -149,14 +149,17 @@ export class StarkRoutingServiceImpl implements StarkRoutingService {
 	public getStateConfigByUrlPath(urlPath: string): StarkStateConfigWithParams | undefined {
 		let targetRoute: StarkStateConfigWithParams | undefined;
 
+		// separate path from hash
+		const [, path = urlPath, hash]: string[] = urlPath.match(/(.*)#(.*)/) || [];
+
 		const matchedState: StateDeclaration[] = this.getStatesConfig().filter((state: StateDeclaration) => {
-			return (<Function>state.$$state)().url && (<Function>state.$$state)().url.exec(urlPath);
+			return (<Function>state.$$state)().url && (<Function>state.$$state)().url.exec(path, undefined, hash);
 		});
 
 		if (matchedState.length) {
 			targetRoute = {
 				state: matchedState[0],
-				paramValues: (<Function>matchedState[0].$$state)().url.exec(urlPath)
+				paramValues: (<Function>matchedState[0].$$state)().url.exec(path, undefined, hash)
 			};
 		}
 
@@ -387,20 +390,18 @@ export class StarkRoutingServiceImpl implements StarkRoutingService {
 
 		// provide custom default error handler
 		// https://ui-router.github.io/ng1/docs/latest/classes/state.stateservice.html#defaulterrorhandler
-		this.$state.defaultErrorHandler(
-			(error: any): void => {
-				let stringError: string;
-				if (error instanceof Rejection) {
-					stringError = error.toString();
-				} else {
-					stringError = String(error);
-				}
-
-				if (!this.knownRejectionCausesRegex.test(stringError)) {
-					this.defaultErrorHandler.handleError(Error(stringError));
-				}
+		this.$state.defaultErrorHandler((error: any): void => {
+			let stringError: string;
+			if (error instanceof Rejection) {
+				stringError = error.toString();
+			} else {
+				stringError = String(error);
 			}
-		);
+
+			if (!this.knownRejectionCausesRegex.test(stringError)) {
+				this.defaultErrorHandler.handleError(Error(stringError));
+			}
+		});
 	}
 
 	/**
--- a/showcase/src/app/app.component.ts
+++ b/showcase/src/app/app.component.ts
@@ -2,6 +2,7 @@
  * Angular 2 decorators and services
  */
 import { Component, Inject, OnInit } from "@angular/core";
+import { BreakpointObserver, BreakpointState } from "@angular/cdk/layout";
 import {
 	STARK_APP_METADATA,
 	STARK_LOGGING_SERVICE,
@@ -13,12 +14,16 @@ import {
 	StarkRoutingService,
 	StarkSessionService,
 	StarkUser,
-	StarkUserService
+	StarkUserService,
+	StarkRoutingTransitionHook,
+	starkAppInitStateName,
+	starkAppExitStateName
 } from "@nationalbankbelgium/stark-core";
 import { STARK_APP_SIDEBAR_SERVICE, StarkAppSidebarService, StarkMenuConfig } from "@nationalbankbelgium/stark-ui";
-import * as moment from "moment";
-import { BreakpointObserver, BreakpointState } from "@angular/cdk/layout";
+import { Transition } from "@uirouter/angular";
+import { StateObject } from "@uirouter/core";
 import { filter } from "rxjs/operators";
+import * as moment from "moment";
 import { APP_MENU_CONFIG } from "./app-menu.config";
 
 /**
@@ -67,6 +72,37 @@ export class AppComponent implements OnInit {
 
 		this.breakpointObserver.observe([this.mediaQueryMdSm]).subscribe((state: BreakpointState) => {
 			this.isMenuModeActive = state.matches;
+		});
+
+		this.routingService.addTransitionHook(
+			StarkRoutingTransitionHook.ON_SUCCESS,
+			{
+				// match any state except the ones that are children of starkAppInit/starkAppExit or the Ui-Router's root state
+				to: (state?: StateObject): boolean => {
+					if (state && typeof state.name !== "undefined") {
+						const regexInitExitStateName: RegExp = new RegExp("(" + starkAppInitStateName + "|" + starkAppExitStateName + ")");
+						return !state.name.match(regexInitExitStateName) && !(state.abstract && state.name === "");
+					} else {
+						return true; // always match
+					}
+				}
+			},
+			(tran: Transition) => {
+				const hash: string = tran.targetState().params()["#"];
+				if (hash) {
+					this.scrollToHash(hash);
+				}
+			}
+		);
+	}
+
+	public scrollToHash(hash: string): void {
+		// Document not loaded on first run yet
+		setTimeout(() => {
+			const element: HTMLElement | null = document.querySelector(`#${hash}`);
+			if (element) {
+				element.scrollIntoView();
+			}
 		});
 	}
 
--- a/showcase/src/app/demo-rbac/pages/authorization-directives/demo-authorization-directives-page.component.html
+++ b/showcase/src/app/demo-rbac/pages/authorization-directives/demo-authorization-directives-page.component.html
@@ -3,6 +3,7 @@
 	<h1 translate>SHOWCASE.DEMO.SHARED.EXAMPLE_VIEWER_LIST</h1>
 	<div fxFlexLayout="column">
 		<example-viewer
+			id="show-on-permission"
 			[extensions]="['HTML', 'TS']"
 			filesPath="rbac-authorization-directives/show-on-permission"
 			exampleTitle="SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.SHOW_ON_PERMISSION.TITLE"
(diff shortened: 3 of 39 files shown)


carlo-nomes:
The PR is not completely finished, but I'm a bit stuck. The Anchors work, when clicked they update the location and scroll into view (adding a hash to the url also scrolls it into view). But when going directly to the url (eg: http://localhost:3000/demo-ui/action-bar#action_bar_alternative) UIRouter removes the hash at some point. But I can't seem to figure out how to resolve this. 😞

christophercr:
I suspect the cause of that issue has to do with our logic to calculate the "target state" when having deep links, such logic is here: https://github.com/NationalBankBelgium/stark/blob/2878029ba57c91bc4636b70d314c5540182dff6e/packages/stark-core/src/modules/session/routes.ts#L88-L102

showcase/src/app/app.component.ts
```diff
@@ -68,6 +70,14 @@ export class AppComponent implements OnInit {
 		this.breakpointObserver.observe([this.mediaQueryMdSm]).subscribe((state: BreakpointState) => {
 			this.isMenuModeActive = state.matches;
 		});
+
+		this.routingService.addTransitionHook(StarkRoutingTransitionHook.ON_SUCCESS, { to: "**" }, (tran: Transition) => {
```
christophercr: The "init" and "exit" states should be excluded here cause there is no reason to do this logic in those.

Something like this:
https://github.com/NationalBankBelgium/stark/blob/master/packages/stark-core/src/modules/session/services/session.service.ts#L150L162

showcase/src/app/shared/components/example-viewer/example-viewer.component.ts
```diff
@@ -22,13 +29,34 @@ export class ExampleViewerComponent implements OnInit {
...
 	@Input()
 	public exampleTitle = "undefined";
+	@Input()
+	public id?: string;
+
+	public get exampleId(): string {
+		return (
+			this.id ||
+			this.translate
+				.instant(this.exampleTitle)
```
christophercr: I don't think is a good idea to use the `exampleTitle` as a fallback in case there is no id, because there might be cases where the title can be the same for 2 or more viewers in the same page.

I suggest to just use the `id` and throw an error if it is not provided... what do you think?

showcase/src/app/shared/shared.module.ts
```diff
@@ -44,7 +45,8 @@ import { FileService } from "./services";
 		MatTooltipModule,
 		MatSnackBarModule,
 		StarkPrettyPrintModule,
-		TranslateModule
+		TranslateModule.forChild(),
```
christophercr: If we remove the `translate.instant` call in the ExampleViewer, then this can be removed too

showcase/src/app/shared/components/example-viewer/example-viewer.component.ts
carlo-nomes: I agree that the title isn't the greatest id, but I wouldn't throw an error if the id is not set. I would rather generated a random id (`example-1823947`) or disable the functionality when there is no id set.

christophercr: Right, then I suggest we go for option 1: generate a random id.

If you want, once you implement the change, I can take over the PR and add the `id` input to every ExampleViewer in the showcase ;)  I just want to be sure we have meaningful/descriptive URLs of every example.

carlo-nomes: I went for option 2 and disabled the anchor link when no id is given. I didn't realise but a random ID won't work. Because the same element will have a different id when visiting the page again, that kind of defeats the purpose I think.

I could make a hash of the title, this at least would not require the translate module being imported.

christophercr: True, random id's won't work. So I go back to my first idea, to throw an error if the id is not given, this will make sure we don't forget to set this id and the implementation will be simple so no Translate module required. To be honest, I don't find any reason why we should make this behaviour optional... it is just used in our showcase so we should be responsible in maintaining this as functional/complete as possible.

As I said, if you don't have time to adapt all the places where we use the `ExampleViewer`, I can do it in this PR once you implement this change.

carlo-nomes: I personally don't like throwing errors for "non-essential" properties and forcing developers to use them (even if it is just us 😂). Maybe sometimes you would not want to add the anchor (you have two examples next to each other and add a title with an anchor or you just have a single example on the page). If we do want a reminder so we don't forget about it I would suggest a warning.
@SuperITMan, @nicanac what do you think I should do?

As for the current examples. I indeed think I won't have time to add them everywhere. 😕 But the PR obviously is still not finished. Because I can't seem to find how to handle landing on an url with a hash part. 😖

showcase/src/app/app.component.ts
carlo-nomes: I'm not sure what you are asking for here. 😕 This needs to be run on every state change right?

christophercr: Well, that's the point, it should run on every state EXCEPT on the "init" and "exit" states because in those there is no content to show...

carlo-nomes: Ah, k I understand. 😄

carlo-nomes: For some reason it broke the functionality.. 🤔 But if I understand correctly if I don't add that code it just gets run 2 times to many during the application lifetime right?

christophercr: Yep, the code will get executed also in the login and the logout pages

showcase/src/app/app.component.ts
```diff
@@ -68,6 +70,21 @@ export class AppComponent implements OnInit {
...
+			setTimeout(() => {
+				const element: HTMLElement | null = document.querySelector(`#${hash}`);
+				if (element) {
+					element.scrollIntoView();
+				}
+			});
+		};
+
+		this.routingService.addTransitionHook(StarkRoutingTransitionHook.ON_FINISH, {}, scrollToHash);
+		this.routingService.addTransitionHook(StarkRoutingTransitionHook.ON_ERROR, {}, scrollToHash);
```
christophercr: I don't think you need the `ON_ERROR` hook here. I know that UI Router calls this hook whenever a navigation is triggered to the same state but I would say that this should be fixed in the `StarkRoutingService` or maybe in the UI Router itself if applicable, so you can simply remove this line ;)

showcase/src/app/shared/components/example-viewer/example-viewer.component.ts
```diff
@@ -22,13 +28,23 @@ export class ExampleViewerComponent implements OnInit {
 	public filesPath = "undefined";
 	@Input()
 	public exampleTitle = "undefined";
+	@Input()
+	public id = "";
+
+	public get exampleState(): string {
+		return this.routingService.getCurrentState().name;
```
christophercr: I know that defining some properties as getters is more concise sometimes, however in this case I think we should use a simple property to keep it consistent with the rest of properties in this class. What do you think?

packages/stark-core/src/modules/routing/services/routing.service.ts
```diff
@@ -149,14 +149,17 @@ export class StarkRoutingServiceImpl implements StarkRoutingService {
 	public getStateConfigByUrlPath(urlPath: string): StarkStateConfigWithParams | undefined {
 		let targetRoute: StarkStateConfigWithParams | undefined;
 
+		// separate path from hash
+		const [, path = urlPath, hash]: string[] = urlPath.match(/(.*)#(.*)/) || [];
+
 		const matchedState: StateDeclaration[] = this.getStatesConfig().filter((state: StateDeclaration) => {
-			return (<Function>state.$$state)().url && (<Function>state.$$state)().url.exec(urlPath);
+			return (<Function>state.$$state)().url && (<Function>state.$$state)().url.exec(path, undefined, hash);
```
christophercr: Can you implement/adapt unit tests for this new functionality?

showcase/src/app/shared/components/example-viewer/example-viewer.component.ts
carlo-nomes: The only problem that could happen is when the state changes during the lifetime of the component. But I suppose that is unlikely

christophercr: Yeah, very very unlikely IMO ;)

showcase/src/app/shared/components/example-viewer/example-viewer.component.html
```diff
@@ -1,7 +1,10 @@
-<mat-card class="stark-card">
+<mat-card [id]="id" class="stark-card">
 	<mat-card-header>
-		<mat-card-title>
-			<h3 translate>{{ exampleTitle }}</h3>
+		<mat-card-title fxLayout fxLayoutAlign="start center">
+			<a *ngIf="id" mat-icon-button [uiSref]="exampleState" [uiParams]="{ '#': id }">
```
christophercr: Can you add a unit test for this new functionality?  Also when there is no `id` set?

showcase/src/app/shared/components/example-viewer/example-viewer.component.ts
```diff
@@ -22,14 +28,23 @@ export class ExampleViewerComponent implements OnInit {
...
 	public showSource = false;
 
-	public constructor(private fileService: FileService, @Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService) {
+	public constructor(
+		private fileService: FileService,
+		@Inject(STARK_LOGGING_SERVICE) public logger: StarkLoggingService,
+		@Inject(STARK_ROUTING_SERVICE) private routingService: StarkRoutingService
+	) {
 		this.appBaseHref = this.getAppBaseHref();
+		this.exampleState = this.routingService.getCurrentStateName();
```
christophercr: Can you also add a unit test for this?

carlo-nomes:
That's ok with me. 😁

showcase/src/app/demo-rbac/pages/authorization-directives/demo-authorization-directives-page.component.html
```diff
@@ -19,11 +20,14 @@ <h1 translate>SHOWCASE.DEMO.SHARED.EXAMPLE_VIEWER_LIST</h1>
 
 			<div *starkShowOnPermission="onShowPermissionsConfig" class="protected-content">
 				{{ "SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.CONTENT_PROTECTED" | translate }}<br />
-				<b>{{ "SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.SHOW_ON_PERMISSION.ONLY_VISIBLE_TO" | translate: { role: "manager" } }}</b>
+				<b>{{
+					"SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.SHOW_ON_PERMISSION.ONLY_VISIBLE_TO" | translate: { role: "manager" }
```
carlo-nomes: Could you put the brackets on the same line as the string here? eg: `{{ "..." }}`

showcase/src/app/demo-rbac/pages/authorization-directives/demo-authorization-directives-page.component.html
```diff
@@ -40,7 +44,9 @@ <h1 translate>SHOWCASE.DEMO.SHARED.EXAMPLE_VIEWER_LIST</h1>
 
 			<div *starkHideOnPermission="onHidePermissionsConfig" class="protected-content">
 				{{ "SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.CONTENT_PROTECTED" | translate }}<br />
-				<b>{{ "SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.HIDE_ON_PERMISSION.ONLY_VISIBLE_TO" | translate: { role: "manager" } }}</b>
+				<b>{{
+					"SHOWCASE.DEMO_RBAC.DIRECTIVES.AUTHORIZATION.HIDE_ON_PERMISSION.ONLY_VISIBLE_TO" | translate: { role: "manager" }
```
carlo-nomes: Same here. 😄

christophercr:
Formatting issues fixed 😉 

@nicanac @SuperITMan could you have a look on this PR?

carlo-nomes:
@SuperITMan That was an confusing rebase. 😅 But it should be done. 😉

christophercr:
And indeed, as @carlo-nomes put in the commit message, it is considered a feature since it is not a demand yet nor a reported bug from anyone.

Issue #1352, opened 2019-07-03
ui: table - Add option to change yPosition of column filter


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Currently when the user opens a column filter, it is opened below the column name. 
Because of that, the first row is partially hidden while the filter is present.

## Expected behavior

The developer should be able to choose the yPosition of the filter (below or above).

## What is the motivation / use case for changing the behavior?

Improve user experience.

## Environment

<pre><code>
Angular version: 7.x
Stark version: 10.0.0-beta.8

</code></pre>

PR #1391, opened 2019-08-26
feat(stark-ui): put the column filter of the table above the title

nicanac:
ISSUES CLOSED: #1352

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In column filtering, there is a little popup screen to type the filter-characters. Now it opens under the column header, which makes it difficult to see the results of the filter.

Issue Number: #1352

## What is the new behavior?
Open it above the column header instead.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/table/components/column.component.ts
+++ b/packages/stark-ui/src/modules/table/components/column.component.ts
@@ -20,7 +20,12 @@ import { FormControl } from "@angular/forms";
 import { distinctUntilChanged } from "rxjs/operators";
 import isEqual from "lodash-es/isEqual";
 import get from "lodash-es/get";
-import { StarkColumnFilterChangedOutput, StarkColumnSortChangedOutput, StarkTableColumnSortingDirection } from "../entities";
+import {
+	StarkColumnFilterChangedOutput,
+	StarkColumnSortChangedOutput,
+	StarkTableColumnFilter,
+	StarkTableColumnSortingDirection
+} from "../entities";
 
 /**
  * Component to display a column inside the StarkTableComponent
@@ -129,6 +134,24 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
 	 */
 	@Input()
 	public sortable = true;
+
+	/**
+	 * Position where the column filter box should be displayed. Default: "below"
+	 */
+	@Input()
+	public get filterPosition(): StarkTableColumnFilter["filterPosition"] {
+		return this._filterPosition;
+	}
+
+	public set filterPosition(value: StarkTableColumnFilter["filterPosition"]) {
+		this._filterPosition = value || "below";
+	}
+
+	/**
+	 * @ignore
+	 * @internal
+	 */
+	private _filterPosition?: StarkTableColumnFilter["filterPosition"] = "below";
 
 	/**
 	 * Priority of the column.
--- a/packages/stark-ui/src/modules/table/components/table.component.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.ts
@@ -34,6 +34,7 @@ import { StarkAction, StarkActionBarConfig } from "../../action-bar/components";
 import {
 	StarkColumnFilterChangedOutput,
 	StarkColumnSortChangedOutput,
+	StarkTableColumnFilter,
 	StarkTableColumnProperties,
 	StarkTableColumnSortingDirection,
 	StarkTableFilter,
@@ -53,7 +54,8 @@ const componentName = "stark-table";
  * Default filter configuration
  */
 const defaultFilter: StarkTableFilter = {
-	globalFilterPresent: true
+	globalFilterPresent: true,
+	filterPosition: "below"
 };
 
 /**
@@ -951,6 +953,20 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 			}
 		}
 		return columnFilterValue;
+	}
+
+	/**
+	 * Get the position of the filter box of the specified column.
+	 * @param columnName - Name of the column whose filter box position should be retrieved.
+	 * @returns The position of the filter box of the specified column.
+	 */
+	public getColumnFilterPosition(columnName: string): StarkTableColumnFilter["filterPosition"] {
+		if (!(this.filter.columns instanceof Array)) {
+			return undefined;
+		}
+
+		const column = this.filter.columns.find((columnFilter: StarkTableColumnFilter) => columnFilter.columnName === columnName);
+		return column ? column.filterPosition : undefined;
 	}
 
 	/**
--- a/packages/stark-ui/src/modules/table/components/column.component.html
+++ b/packages/stark-ui/src/modules/table/components/column.component.html
@@ -16,7 +16,7 @@
 				<mat-icon class="stark-small-icon" svgIcon="filter" [matTooltip]="'STARK.TABLE.COLUMN_FILTER' | translate"></mat-icon>
 			</button>
 		</div>
-		<mat-menu class="mat-table-filter" #filterMenu="matMenu" xPosition="before" [overlapTrigger]="false">
+		<mat-menu class="mat-table-filter" #filterMenu="matMenu" [yPosition]="filterPosition" xPosition="before" [overlapTrigger]="false">
 			<div>
 				<mat-form-field
 					class=""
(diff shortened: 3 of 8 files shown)


christophercr:
Btw, @nicanac could you adapt a bit the commit message to mention that this change is about the Table?

Keep in mind that the commit messages are put in the `CHANGELOG` so they need to be descriptive enough for the developers using our framework ;)

nicanac:
> This change would put the "global" filter above the table which I think it shouldn't be put there:
> global filter above
> 
> But in fact, I think the issue #1352 is about the column filter which overlaps with the first row:
> column filter below
> 
> @SuperITMan can you confirm this?

Yes, that's right, I've update the pr to move the filter of the column and not the global filter.
Nonetheless, isn't strange having the global filter above and column filter below ?
@SuperITMan, @christophercr, What do you think ?

nicanac:
> Btw, @nicanac could you adapt a bit the commit message to mention that this change is about the Table?
> 
> Keep in mind that the commit messages are put in the `CHANGELOG` so they need to be descriptive enough for the developers using our framework ;)

@christophercr , mention table added.

carlo-nomes:
I would make both configurable by the developer and keep the defaults like they currently are (below the button) because that's the least likely direction for it to be rendered outside the viewport.
For the column filter... from a UX standpoint, and if it would be build from the ground up I would suggest doing something like [this](https://codepen.io/carlo-nomes/pen/BaBRKZO) that way no important screen real estate is lost. But that's probably too much work for this PR.

nicanac:
> I would make both configurable by the developer and keep the defaults like they currently are (below the button) because that's the least likely direction for it to be rendered outside the viewport.
> For the column filter... from a UX standpoint, and if it would be build from the ground up I would suggest doing something like [this](https://codepen.io/carlo-nomes/pen/BaBRKZO) that way no important screen real estate is lost. But that's probably too much work for this PR.

For now, I've added the possibility to display the filter above or below. Default is below.

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -130,6 +130,12 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
 	@Input()
 	public sortable = true;
 
+	/**
+	 * Display the column filter box below or above. Default: below
+	 */
+	@Input()
+	public searchFilterYPosition? = "below";
```
christophercr: This should not be a separate `input` on the component, but instead it should be put together with all the properties of the column filter here: [packages/stark-ui/src/modules/table/entities/table-column-filter.intf.ts](https://github.com/NationalBankBelgium/stark/blob/master/packages/stark-ui/src/modules/table/entities/table-column-filter.intf.ts)

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -168,6 +169,12 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	@Input()
 	public fixedHeader?: string;
 
+	/**
+	 * YPosition of the searchFilter box for the table and the column
+	 */
+	@Input()
+	public searchFilterYPosition?: "above" | "below" = "below";
```
christophercr: This should not be a separate input on the component, but instead it should be put together with all the properties of the table filter here: [/packages/stark-ui/src/modules/table/entities/table-filter.intf.ts](https://github.com/NationalBankBelgium/stark/blob/master/packages/stark-ui/src/modules/table/entities/table-filter.intf.ts)

packages/stark-ui/src/modules/table/components/column.component.ts
nicanac: ah ok, good to know, it's a lot more clean.

nicanac:
> Maybe we could base this functionality on the [MatTooltip TooltipPosition](https://material.angular.io/components/tooltip/api#TooltipPosition). I would not expand this PR to include the `"left"`, `"right"`, ... now. But I think it would be nice to future proof this input and rename it to `position` so we could add this functionality later without introducing a breaking change. What do you think @christophercr , @nicanac ?

Yes, sure it's a good idea.
But as you said, I think it'll be easier to do it in a separated issue. 
What's your opinion @christophercr ?

nicanac:
> > LGTM 👍
> > @carlo-nomes @SuperITMan could you guys also review this PR?
> 
> Passing the test, it seems I've an issue with the globalFilterValue

I've change a bit the test, and it doesn't fail anymore.

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -953,6 +955,20 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 		return columnFilterValue;
 	}
 
+	/**
+	 * Get the position of the filter box of the specified column.
+	 * @param columnName - Name of the column whose filter box position should be retrieved.
+	 * @returns The position of the filter box of the specified column.
+	 */
+	public getColumnFilterPosition(columnName: string): StarkTableColumnFilter["filterPosition"] | undefined {
```
christophercr: Here there is no need to add `| undefined` because it is already part of the types from the `filterPosition` property.

Can you remove it?

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -953,6 +955,20 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
+		if (!(this.filter.columns instanceof Array)) {
+			return;
```
christophercr: This should return something explicitly, in this case `undefined`

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -953,6 +955,20 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
+		}
+
+		const column = this.filter.columns.find((columnFilter: StarkTableColumnFilter) => columnFilter.columnName === columnName);
+		return column && column.filterPosition;
```
christophercr: What is returned here if there is no column? `null`? `undefined`? 

I think we should return something explicitly: `undefined`

packages/stark-ui/src/modules/table/components/table.component.ts
carlo-nomes: I think TS was complaining when it wasn't added while I was testing. 🤔

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: yes that's right, I've adapted the code

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: yes that's right, I've adapted the code

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: I removed it, seems TS is fine with it :-)

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: Does it mean we need a new if case here ?

christophercr: Mmm yes, but you can do it in the same line like this:

`return column ? column.filterPosition : undefined;`

nicanac: hm, ok I see.

It's changed :-)

packages/stark-ui/src/modules/table/components/table.component.ts
carlo-nomes: K, great. 😄

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -953,6 +955,20 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
...
+	 * @param columnName - Name of the column whose filter box position should be retrieved.
+	 * @returns The position of the filter box of the specified column.
+	 */
+	public getColumnFilterPosition(columnName: string): StarkTableColumnFilter["filterPosition"] {
+		if (!(this.filter.columns instanceof Array)) {
+			return undefined;
+		}
+
+		const column = this.filter.columns.find((columnFilter: StarkTableColumnFilter) => columnFilter.columnName === columnName);
+		return column && column.filterPosition;
```
christophercr: This line still needs to fixed ;)

`return column ? column.filterPosition : undefined;`

PR #1560, opened 2020-01-21
fix(build-main): fix sync scripts for starter + add 'showcase' in scopes of commitlint

SuperITMan:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[X] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The "sync" scripts in the root "package.json" are wrong for syncing with the "starter".

When we run `npm run commit`, we cannot choose the scope "showcase" nor "stark-showcase". This scope is missing.

## What is the new behavior?

Fix the "sync" scripts.

Change the configuration of commitlint. The scope "stark-starter" is now "starter". The scope "showcase" has been added. 

I think "starter" and "showcase" is clear enough for the scopes and the "stark-" prefix is not necessary. What do you think @nicanac @christophercr ? :blush:  

```
[ ] Yes
[X] No
```


CONTRIBUTING.md
```diff
@@ -446,9 +446,10 @@ Must be one of the following:
 -   **stark-core**
 -   **stark-demo**
 -   **stark-rbac**
--   **stark-starter**
 -   **stark-testing**
 -   **stark-ui**
+-   **showcase**
+-   **starter**
```
christophercr: Not sure whether we should change these scopes: we already have `stark-demo` (aka showcase) and `stark-starter` (aka starter) and we have used them frequently (check the [CHANGELOG](https://github.com/NationalBankBelgium/stark/blob/master/CHANGELOG.md)). The only one that we have rarely used and it was not defined is `showcase`.

I have no strong opinion on this, just the fact that we will have several names for these in our changelog. On the other hand, it is true that we haven't reached a final version so we can technically change the naming easily.

If we change the naming, then I agree that the `stark-` is not relevant and we should also remove `stark-demo` from the scopes too.

SuperITMan: As you say, we're not yet in final release. I think having a shorter name for those two should be better.
Good catch for the "stark-demo" ! 😊 (which is actually absent in "commitlint.config.js").

@nicanac What do you think ? Do you validate the following replacements ?
  - stark-demo ➡️ showcase
  - stark-starter ➡️ starter

nicanac: Yes sure, my only question is, are we going to rewrite the changelog ?

SuperITMan: Sorry, I didn't see your comment and I added mine... 
Actually, it concerns the starter and the showcase, it's linked to Stark itself. 
Because of that, I think it's not a big deal if the CHANGELOG has different values for "showcase", "stark-demo", "stark-starter" and "starter". 
Since the CHANGELOG is generated automatically based on the commit messages, we won't change it 😊

commitlint.config.js
```diff
@@ -20,9 +20,10 @@ module.exports = {
 				"stark-core",
 				"stark-demo",
```
christophercr: This one should be removed too right?

SuperITMan: Yes, you're right. 
Btw, shouldn't we also remove the "build" or the "build-main" ?
I have the feeling those 2 scopes are for the same thing 😕 

What do you think @nicanac @christophercr ?

christophercr: Yep, I think they are referring to the same thing... so maybe we can just keep "build-main"? Because that one is more specific and can't be easily overlooked than "build" which could be confused with "stark-build". Moreover, we already have a "build-main" label in the Github repo ;)

SuperITMan:
@nicanac @christophercr PR updated :blush:

PR #1573, opened 2020-01-24
Update lint-staged configuration in root & starter + remove lint-staged from showcase

SuperITMan:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When we run `npm install && npm run install:all`, the husky hooks are not well configured because the "starter" is installed after the root. 
The result is when we do a commit, only the git hook set in the "starter" is triggered instead of the one configured in the root folder.

Also, the showcase contains the dependency "lint-staged" even if it is not needed neither used.

The husky config is not complete in the starter which means that no hook is actually triggered.

Issue Number: N/A

## What is the new behavior?

Updated lint-staged configuration in the root and in the starter, removing the useless "git add ." which is done automatically. 

When we run `npm run install:starter`, the "husky" dependency is not installed thanks to the variable "SKIP_HUSKY_INSTALL". 

Removed the useless "lint-staged" dependency from the "showcase".

```
[ ] Yes
[X] No
```

## Other information


starter/package.json
```diff
@@ -62,7 +62,6 @@
     "postinstall": "npm run webdriver:update",
     "postversion": "git push && git push --tags",
     "preclean:install": "npm run clean",
-    "precommit": "lint-staged",
```
christophercr: Why is this removed in the starter if `lint-staged` is still in the `devDependencies` list?

SuperITMan: Because this script is useless.
It is not present in our main package.json neither

christophercr: Ooops... I missed that, sorry :p

Issue #1540, opened 2020-01-16
ui: table implement footer


nicanac:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Actually, there's no table summering a column in a footer available for the developers.

## Expected behavior

As it's available in angular material, we could simply add it to stark.
Angular Material: https://material.angular.io/components/table/examples
Example: https://stackblitz.com/angular/yarvnglavmk?file=src%2Fapp%2Ftable-footer-row-example.ts

PR #1601, opened 2020-01-30
feat(stark-ui): add footer to table component

nicanac:
ISSUES CLOSED: #1540

## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#-commit-message-guidelines

## PR Type
Implement a footer for the table component

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1540 

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/stark-ui/src/modules/table/components/column.component.html
+++ b/packages/stark-ui/src/modules/table/components/column.component.html
@@ -39,4 +39,9 @@
 			*ngTemplateOutlet="columnTemplate; context: { $implicit: { rowData: rowItem, displayedValue: getDisplayedValue(rowItem) } }"
 		></ng-container>
 	</td>
+	<ng-container *ngIf="footerValue !== 'undefined'">
+		<td mat-footer-cell *matFooterCellDef>
+			{{ footerValue!.toString() | translate }}
+		</td>
+	</ng-container>
 </ng-container>
--- a/packages/stark-ui/src/modules/table/components/column.component.ts
+++ b/packages/stark-ui/src/modules/table/components/column.component.ts
@@ -131,6 +131,27 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
 	private _headerLabel?: string;
 
 	/**
+	 * Value to be shown as the column's footer.
+	 */
+	@Input()
+	public set footerValue(value: string | number | undefined) {
+		this._footerValue = value;
+	}
+
+	/**
+	 * Returns the footer value of the column if it's specified.
+	 */
+	public get footerValue(): string | number | undefined {
+		return this._footerValue;
+	}
+
+	/**
+	 * @ignore
+	 * @internal
+	 */
+	private _footerValue?: string | number;
+
+	/**
 	 * Whether the column is sortable or not. Default: true
 	 */
 	@Input()
--- a/packages/stark-ui/src/modules/table/components/table.component.ts
+++ b/packages/stark-ui/src/modules/table/components/table.component.ts
@@ -45,6 +45,7 @@ import { AbstractStarkUiComponent } from "../../../common/classes/abstract-compo
 import { StarkPaginateEvent, StarkPaginationComponent, StarkPaginationConfig } from "../../pagination/components";
 import { StarkMinimapComponentMode, StarkMinimapItemProperties } from "../../minimap/components";
 import find from "lodash-es/find";
+import findIndex from "lodash-es/findIndex";
 
 /**
  * Name of the component
@@ -68,6 +69,8 @@ const DEFAULT_COLUMN_PROPERTIES: Partial<StarkTableColumnProperties> = {
 	isVisible: true
 };
 
+// FIXME: refactor the template of this component function to reduce its cyclomatic complexity
+/* tslint:disable:template-cyclomatic-complexity */
 /**
  * Component to display array data in a table layout.
  */
@@ -82,16 +85,22 @@ const DEFAULT_COLUMN_PROPERTIES: Partial<StarkTableColumnProperties> = {
 		class: componentName
 	}
 })
-/* tslint:enable */
 export class StarkTableComponent extends AbstractStarkUiComponent implements OnInit, AfterViewInit, OnChanges, OnDestroy {
 	/**
 	 * Array of {@link StarkTableColumnProperties} objects which define the columns of the data table.
 	 */
 	@Input()
 	public set columnProperties(input: StarkTableColumnProperties[]) {
+		this.isFooterEnabled =
+			findIndex(
+				input || [],
+				(column: StarkTableColumnProperties) => typeof column.footerValue !== "undefined" && column.footerValue !== ""
+			) > -1;
+
 		this._columnProperties = (input || []).map((properties: StarkTableColumnProperties) => ({
 			...DEFAULT_COLUMN_PROPERTIES,
-			...properties
+			...properties,
+			footerValue: this.isFooterEnabled ? properties.footerValue || "" : undefined
 		}));
 
 		if (this.dataSource) {
@@ -423,6 +432,14 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	 * Whether the fixed header is enabled.
 	 */
 	public isFixedHeaderEnabled = false;
+
+	/**
+	 * Whether the footer is enabled.
+	 * Default: if there is no footer defined here and in any other column, then it won't be displayed.
+	 * Otherwise, if at least one of the other columns defines a footer,
+	 * then the footer of this column will be displayed as empty
+	 */
+	public isFooterEnabled = false;
 
 	/**
 	 * Whether the current sorting is done on multiple columns
--- a/packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
+++ b/packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
@@ -61,6 +61,14 @@ export interface StarkTableColumnProperties {
 	label?: string;
 
 	/**
+	 * Value to be shown as the column's footer.
+	 * Default: if there is no footer defined here and in any other column,
+	 * then it won't be displayed. Otherwise, if at least one of the other columns defines a footer,
+	 * then the footer of this column will be displayed as empty
+	 */
+	footerValue?: string | number;
+
+	/**
 	 * Name of the property that will be the source of the column.
 	 */
 	name: string;
--- a/showcase/src/app/demo-ui/components/table-with-footer/table-with-footer.component.ts
+++ b/showcase/src/app/demo-ui/components/table-with-footer/table-with-footer.component.ts
@@ -1,0 +1,48 @@
+import { Component, ViewEncapsulation } from "@angular/core";
+import { StarkPaginationConfig, StarkTableColumnProperties, StarkTableFilter } from "@nationalbankbelgium/stark-ui";
+
+const DUMMY_DATA: object[] = [
+	{ id: 1, cost: 12, description: "number one" },
+	{ id: 10, cost: 23, description: "second description" },
+	{ id: 12, cost: 5, description: "the third description" },
+	{ id: 2, cost: 33, description: "description number four" },
+	{ id: 23, cost: 54, description: "fifth description" },
+	{ id: 222, cost: 3, description: "the sixth description" },
+	{ id: 112, cost: 7, description: "seventh description" },
+	{ id: 232, cost: 24, description: "description number eight" },
+	{ id: 154, cost: 35, description: "the ninth description" },
+	{ id: 27, cost: 10, description: "description number ten" },
+	{ id: 86, cost: 21, description: "eleventh description" },
+	{ id: 44, cost: 6, description: "the twelfth description" }
+];
+
+@Component({
+	selector: "showcase-table-with-footer",
+	templateUrl: "./table-with-footer.component.html",
+	styleUrls: ["./table-with-footer.component.scss"],
+	/* tslint:disable-next-line:use-view-encapsulation */
+	encapsulation: ViewEncapsulation.None // Important
+})
+export class TableWithFooterComponent {
+	public data: object[] = DUMMY_DATA;
+	public paginationConfig: StarkPaginationConfig = {
+		itemsPerPage: 10
+	};
+
+	public columns: StarkTableColumnProperties[] = [
+		{ name: "id", label: "Id", footerValue: "SHOWCASE.DEMO.TABLE.FOOTER.TOTAL" },
+		{
+			name: "cost",
+			label: "SHOWCASE.DEMO.TABLE.LABELS.TITLE",
+			footerValue: this.getTotal()
+		},
+		{ name: "description", label: "SHOWCASE.DEMO.TABLE.LABELS.DESCRIPTION" }
+	];
+
+	public filter: StarkTableFilter = { globalFilterPresent: false, columns: [] };
+
+	/** Gets the total cost of all products. */
+	private getTotal(): number {
+		return DUMMY_DATA.map((obj: any) => obj.cost).reduce((acc: number, value: number) => acc + value, 0);
+	}
+}
--- a/showcase/src/assets/examples/table/with-footer/table.html
+++ b/showcase/src/assets/examples/table/with-footer/table.html
@@ -1,0 +1,3 @@
+<stark-table [data]="data" [columnProperties]="columns" [paginationConfig]="paginationConfig" [filter]="filter">
+	<header><h1 class="mb0">Table with footer</h1></header>
+</stark-table>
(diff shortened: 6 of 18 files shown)


showcase/src/assets/examples/table/with-footer/table.ts
```diff
@@ -0,0 +1,48 @@
+import { Component, ViewEncapsulation } from "@angular/core";
+import { StarkPaginationConfig, StarkTableColumnProperties, StarkTableFilter } from "@nationalbankbelgium/stark-ui";
+
+const DUMMY_DATA: object[] = [
```
SuperITMan: Could you please reduce this object as we did in another example: https://github.com/NationalBankBelgium/stark/blob/52ae6e0380e24bd3678f0df15edd53746c6ac79c/showcase/src/assets/examples/table/with-custom-styling/table.ts#L4 

```typescript
const DUMMY_DATA: object[] = [
	{ id: 1, cost: 12, description: "number one" },
        /* ... */
       	{ id: 44, cost: 6, description: "the twelfth description" }
];
```

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
```diff
@@ -60,6 +60,11 @@ export interface StarkTableColumnProperties {
 	 */
 	label?: string;
 
+	/**
+	 * Value to be shown as the column's footer.
+	 */
+	footerValue?: any;
```
SuperITMan: Shouldn't it be a `string` here ?

packages/stark-ui/src/modules/table/components/column.component.html
```diff
@@ -39,4 +39,7 @@
 			*ngTemplateOutlet="columnTemplate; context: { $implicit: { rowData: rowItem, displayedValue: getDisplayedValue(rowItem) } }"
 		></ng-container>
 	</td>
+	<td mat-footer-cell *matFooterCellDef>
+		<span>{{ footerValue }}</span>
```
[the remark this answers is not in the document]
nicanac: Actually, I took the inspiration from the table.component 			
`<ng-template let-context>
				<span>{{ context.displayedValue }}</span>
			</ng-template>`

But you're right, i's a bit useless, I'll remove it

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
nicanac: I don't know, do you think it'll alway be  a string. I think it can be a number too ?

showcase/src/assets/examples/table/with-footer/table.ts
nicanac: Yes, forget it :-), thanks

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
SuperITMan: Mmm.. You're right! 😊 
Let's keep it like this for now

nicanac: oki ok

christophercr: In fact I think we should to set `string|number` to make our API clear enough... later on we might change this value to anything as @nicanac suggests.

Also, I think we should add the following to the JSDoc: `Default: if there is no footer defined here and in any other column, then it won't be displayed. Otherwise, if at least one of the other columns defines a footer, then the footer of this culumn will be displayed as empty`. What do you think?

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -68,6 +68,8 @@ const DEFAULT_COLUMN_PROPERTIES: Partial<StarkTableColumnProperties> = {
 	isVisible: true
 };
 
+// FIXME: refactor the template of this component function to reduce its cyclomatic complexity
+/* tslint:disable:template-cyclomatic-complexity */
```
christophercr: Can you move these lines to the actual function that needs to be fixed? That way it is easier to spot where the problem is ;)

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -171,6 +172,12 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	@Input()
 	public fixedHeader?: string;
 
+	/**
+	 * Determine if the footer will be displayed or not
+	 */
+	@Input()
+	public footer?: boolean;
```
christophercr: I don't think this input is needed because it will only depend on the `footer` property of the `ColumnProperties` right?

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -130,6 +130,27 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
 	 */
 	private _headerLabel?: string;
 
+	/**
+	 * Value to be shown as the column's footer.
+	 */
+	@Input()
+	public set footerValue(value: any) {
```
christophercr: Same here: use `string|number` instead of `any` ;)

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -424,6 +431,11 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 	 */
 	public isFixedHeaderEnabled = false;
 
+	/**
+	 * Whether the footer is enabled.
+	 */
+	public isFooterEnabled = false;
```
christophercr: So the value of this property should be disabled only if none of the columns define a footer...

packages/stark-ui/src/modules/table/components/column.component.ts
nicanac: Yeah at first It was what I used, but I was afraid to be to strict. Maybe a day we'll need an object. But I agree, for now I'll put it as string|number

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
nicanac: Yes, it was a bit poor in doc, thanks

nicanac: and also, I've a remark of tslint when I use string|number , it also requires the value undefined.
Is it ok for you

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: @christophercr, actually the issue is in the html file, is it possible to use it there ?

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: Actually, I use it here line 556
`		if (changes["footer"]) {
			this.isFooterEnabled = coerceBooleanProperty(this.footer);
		}`

and then I tested in the html of the table to display the line : 
`		<ng-container *ngIf="isFooterEnabled">
			<tr mat-footer-row *matFooterRowDef="displayedColumns"></tr>
		</ng-container>`

Maybe it's not the best way to do it ?

packages/stark-ui/src/modules/table/components/table.component.ts
christophercr: Oops, my bad.. I didn't read the `template` part in the rule name... so no worries, it is ok to leave it where you put it ;)

packages/stark-ui/src/modules/table/components/table.component.ts
christophercr: But as I explained in my other remark, the footer should be shown based on what the developer passes in the `ColumnProperties`...

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: oki ok nice

packages/stark-ui/src/modules/table/components/column.component.ts
christophercr: Indeed... cause maybe later on we'll need to implement a formatter or something to display more complex values...

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
christophercr: but I think `footerValue?: string|number;` already covers the `undefined` type right? Or am I missing something?

nicanac: image
image

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: Ok, I'll get, I'll modify to get it from the column.

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
christophercr: ah ok, the getter/setter... yeah, in that case you should add `|undefined`

packages/stark-ui/src/modules/table/components/column.component.html
```diff
@@ -39,4 +39,7 @@
 			*ngTemplateOutlet="columnTemplate; context: { $implicit: { rowData: rowItem, displayedValue: getDisplayedValue(rowItem) } }"
 		></ng-container>
 	</td>
+	<td mat-footer-cell *matFooterCellDef>
```
christophercr: Shoudln't this footer be added only if footerValue is not `undefined`?

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -94,6 +96,9 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 			...properties
 		}));
 
+		this.isFooterEnabled =
+			findIndex(this.columnProperties, (column: StarkTableColumnProperties) => typeof column.footerValue !== "undefined") > -1;
```
christophercr: `typeof column.footerValue !== "undefined" && column.footerValue !== ''`

To not take into account empty footers

packages/stark-ui/src/modules/table/components/table.component.ts
```diff
@@ -94,6 +96,9 @@ export class StarkTableComponent extends AbstractStarkUiComponent implements OnI
 			...properties
```
christophercr: I think this should be moved after the next line of code (`this.isFooterEnabled = xxxx`).

That way this can be adapted like this:

```typescript
this._columnProperties = (input || []).map((properties: StarkTableColumnProperties) => ({
    ...DEFAULT_COLUMN_PROPERTIES,
    ...properties,
    footerValue: this.isFooterEnabled ? (properties.footerColumn || '') : undefined
}));
```

So in case the footer is enabled, the footerValue of the current column is set to whatever the devevloper passed to it or '' so that it is displayed in the column

packages/stark-ui/src/modules/table/components/column.component.html
nicanac: yes true

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: oki ok

packages/stark-ui/src/modules/table/components/table.component.ts
nicanac: mhhh I did try, but since then I haven't been able to get the footer working anymore.

christophercr: In the code I wrote above I have a typo: `properties.footerColumn` but it should be `properties.footerValue` :p

nicanac: Hi @christophercr,
I think I know why it's not working, if I put the code above under 
`		this.isFooterEnabled =
			findIndex(
				this.columnProperties,
				(column: StarkTableColumnProperties) => typeof column.footerValue !== "undefined" && column.footerValue !== ""
			) > -1;`
this.`columnProperties`, requiers - >`this._columnProperties` set later in the code line 117
`	public get columnProperties(): StarkTableColumnProperties[] {
		return this._columnProperties;
	}`

Maybe I'm wrong ? Any idea how i can solve this ?

packages/stark-ui/src/modules/table/components/column.component.html
```diff
@@ -39,4 +39,9 @@
 			*ngTemplateOutlet="columnTemplate; context: { $implicit: { rowData: rowItem, displayedValue: getDisplayedValue(rowItem) } }"
 		></ng-container>
 	</td>
+	<ng-container *ngIf="footerValue !== 'undefined' && footerValue !== ''">
```
christophercr: In this specific case, we should check only for `undefined`. Because we might set the `footerValue` to `''` if the footer should be enabled because the developer set the `footerValue` for other column..

packages/stark-ui/src/modules/table/components/table.component.ts
christophercr: Oh right... well then I think it is still possible to move it above but the `isFooterEnabled` should be adapted like this:

```typescript
		this.isFooterEnabled =
			findIndex(
				(input || []),
				(column: StarkTableColumnProperties) => typeof column.footerValue !== "undefined" && column.footerValue !== ""
			) > -1;
```

Basically just replace `this.columnProperties` by `(input || [])` which is the same we do in the code above

Tip: you could replace also `findIndex()` by the array [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) function 😉

packages/stark-ui/src/modules/table/components/column.component.html
nicanac: Ok ok I see

showcase/src/assets/examples/table/with-footer/table.html
```diff
@@ -0,0 +1,3 @@
+<stark-table [data]="data" [columnProperties]="columns" [paginationConfig]="paginationConfig" [filter]="filter">
+	<header><h1 class="mb0" translate>Table with footer</h1></header>
```
christophercr: Remove the `translate` directive from this example ;)

showcase/src/assets/examples/table/with-footer/table.ts
```diff
@@ -0,0 +1,39 @@
...
+	{ id: 1, cost: 12, description: "number one" },
+	/* ... */
+	{ id: 44, cost: 6, description: "the twelfth description" }
+];
+
+@Component({
+	selector: "showcase-table-with-footer",
+	templateUrl: "./table-with-footer.component.html",
+	styleUrls: ["./table-with-footer.component.scss"],
+	/* tslint:disable-next-line:use-view-encapsulation */
```
christophercr: Remove this line as it is not relevant for the example code

showcase/src/assets/examples/table/with-footer/table.html
nicanac: Done

showcase/src/assets/examples/table/with-footer/table.ts
nicanac: done

nicanac: > Great! It is working now! Just small remarks and also... don't forget to write unit tests for this new feature 😉

Yes sure, I'll create an issue for this purpose : https://github.com/NationalBankBelgium/stark/issues/1613

packages/stark-ui/src/modules/table/components/column.component.html
```diff
@@ -39,4 +39,9 @@
 			*ngTemplateOutlet="columnTemplate; context: { $implicit: { rowData: rowItem, displayedValue: getDisplayedValue(rowItem) } }"
 		></ng-container>
 	</td>
+	<ng-container *ngIf="footerValue !== 'undefined'">
+		<td mat-footer-cell *matFooterCellDef>
+			{{ footerValue }}
```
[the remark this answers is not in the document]
nicanac: mhhh why not, but if we pass a number, isn't it an issue ?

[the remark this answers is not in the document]
SuperITMan: Mmm... You're right! I thought the pipe was able to get a string but it's not the case.
What we could do to make this working is:
```html
<td mat-footer-cell *matFooterCellDef>
    {{ footerValue!.toString() | translate }}
</td>
```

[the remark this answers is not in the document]
nicanac: oki ok it's implemented.

showcase/src/app/demo-ui/components/table-with-footer/table-with-footer.component.ts
```diff
@@ -0,0 +1,48 @@
...
+	encapsulation: ViewEncapsulation.None // Important
+})
+export class TableWithFooterComponent {
+	public data: object[] = DUMMY_DATA;
+	public paginationConfig: StarkPaginationConfig = {
+		itemsPerPage: 10
+	};
+
+	public columns: StarkTableColumnProperties[] = [
+		{ name: "id", label: "Id", footerValue: "Total" },
```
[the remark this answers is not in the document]
nicanac: done

Issue #1570, opened 2020-01-22
ui: session - dropdown/menu opened in header are not closed when user is logged out


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

When the session is closed by the StarkSession service, it redirects to the session-logout page.
Unfortunately, if the user opened the language selector before being logged out, it stays opened.

![image](https://user-images.githubusercontent.com/10333565/72900054-1b542000-3d1f-11ea-8eee-208168e484a5.png)

## Expected behavior

The dropdown should be closed when the user is redirected to the session-logout page.

## Minimal reproduction of the problem with instructions

Open the language-selector and wait until you are logged out automatically.

## What is the motivation / use case for changing the behavior?

Bug fixing.

## Environment

<pre><code>
Angular version: 7.x
Stark version: 10.0.0-rc.3

</code></pre>

christophercr:
After investigating the issue, it turns out the cause of the problem is that any Angular Material element that creates an overlay, such as the `mat-select` inside the `StarkDropdown` component, is embedded in an [`OverlayContainer`](https://v7.material.angular.io/cdk/overlay/api#OverlayContainer) and this container is not placed inside the `app` component but at the root of the page.

Therefore, when the view is destroyed by the router when navigating to the `SessionExpiredPage`, the container remains there with the overlays that it had at the moment (open or closed):

<img width="1008" alt="Screenshot 2020-03-24 at 10 43 12" src="https://user-images.githubusercontent.com/5698318/77411219-5a086400-6dbc-11ea-97c1-b15f6f8122ae.png">


So, the solution would be then to destroy such `OverlayContainer` when navigating to any of the `exit` states: `SessionExpired` and `SessionLogout`. This can be done by calling its [`ngOnDestroy`](https://v7.material.angular.io/cdk/overlay/api#OverlayContainer) method.

PR #1761, opened 2020-03-24
fix(stark-ui): remove all overlays by destroying the Angular CDK OverlayContainer when navigating to an 'exit' state

christophercr:
ISSUES CLOSED: #1570

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1570 

## What is the new behavior?
All overlays created by the `mat-select` component from the `StarkDropdown` and `mat-menu` from the `StarkActionBar` and other star components are now removed when navigating to an exit state (`StarkSessionEspired` and `StarkSessionLogout`) by destroying the container where all of them are rendered

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/stark-ui/src/modules/session-ui/routes.ts
+++ b/packages/stark-ui/src/modules/session-ui/routes.ts
@@ -1,4 +1,6 @@
+import { HookResult, Transition } from "@uirouter/core";
 import { Ng2StateDeclaration } from "@uirouter/angular";
+import { OverlayContainer } from "@angular/cdk/overlay";
 import {
 	starkLoginStateName,
 	starkLoginStateUrl,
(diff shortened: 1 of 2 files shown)


packages/stark-ui/src/modules/session-ui/routes.ts
```diff
@@ -16,6 +18,20 @@ import {
...
+ * Hook to destroy the OverlayContainer inside which all overlays are rendered (i.e. stark-dropdown options)
+ * Fixes https://github.com/NationalBankBelgium/stark/issues/1570
+ */
+export function destroyOverlaysOnEnterFn(transition: Transition): HookResult {
+	// inject the OverlayContainer
+	const overlayContainer: OverlayContainer = transition.injector().get<OverlayContainer>(OverlayContainer);
+	// destroy the container by calling its own "ngOnDestroy" method
+	// see https://github.com/angular/components/pull/5378/files
+	/* tslint:disable-next-line:no-lifecycle-call*/
+	overlayContainer.ngOnDestroy();
```
SuperITMan: What would happen if no instance of `OverlayContainer` is available because the application does not use it ?

I don't have a good example in mind but wouldn't it be better to put this in a `if` condition ? 😊 
```js
if (typeof overlayContainer !== "undefined") {
    overlayContainer.ngOnDestroy();
}
```

christophercr: Good point... indeed, that might happen if the app don't use any of the stark components that create an overlay (dropdown, app-data, menu, etc).

I'll add the check 😉

christophercr:
@SuperITMan updated this PR with your remark 😉 
Can you have a look?

packages/stark-ui/src/modules/session-ui/routes.ts
SuperITMan: Mmm... I have a small doubt now 😛 
Did you already put the try/catch before?

Because with the try/catch, I'm not sure the `if` is really necessary.
If it was there before, my bad I didn't see it.

christophercr: The `if` wasn't there before, but I added it together with the `try ... catch` just in case we get something injected  but `undefined`. I know it doesn't make not much sense... but I added it just in case... to be on the safe side...

christophercr: In the end I removed the `if` check cause it really doesn't make sense... the `try...catch` is more than enough ;)

Issue #1700, opened 2020-03-06
ui: Table - Allow usage of tooltip


nicanac:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Add the possibility to add a tooltip on a cell value. cfr. screenshot in attachment.

## Expected behavior

Stark Table should provide the possibility to add a tooltip on a cell value.

## What is the motivation / use case for changing the behavior?

Improvement

## Other

Links that could help us:
- [Material doc](https://material.angular.io/components/tooltip/overview)
- [Caution with the number of tooltip used](https://stackoverflow.com/questions/56986405/large-number-of-mat-tooltip-inside-mat-table-slows-application)

PR #2146, opened 2020-09-04
feat(start-ui): allow usage of tooltip

dm148:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No possibility for tooltips in table

Issue Number: #1700 

## What is the new behavior?
You can add tooltip for each cell, you can set the value in StarkTableColumnProperties or in data.
Also possibility to add a class for styling the tooltip.
You can set the position of the tooltip.

```
[ ] Yes
[x] No
```

## Other information


packages/stark-ui/src/modules/table/components/column.component.html
```diff
@@ -34,7 +34,8 @@
 	</th>
 	<!-- the column template defined by the user will be displayed here -->
 	<!-- and it will receive the right context containing the displayedValue and the row data-->
-	<td mat-cell *matCellDef="let rowItem" [ngClass]="getCellClassNames(rowItem)" (click)="onCellClick($event, rowItem)">
+	<td mat-cell *matCellDef="let rowItem" [ngClass]="getCellClassNames(rowItem)" (click)="onCellClick($event, rowItem)" 
+		[matTooltip]="getCellTooltip(rowItem)" [matTooltipPosition]="cellTooltipPosition" [matTooltipClass]="getCellTooltipClassNames(rowItem)">
```
christophercr, line 38: What if the tooltip is an empty string `""`? I think we should disable the tooltip in such cases with the `matTooltipDisabled` property otherwise I fear that Angular Material would still be doing some logic to display the tooltip which will be overkill if that would be an empty string in the end.

Another solution would be to have 2 `<td mat-cell>` with an `*ngIf`:  one will contain the `[matTooltip]` if there is a relevant tooltip to show (not empty). And the other one won't contain the `[matTooltip]`, basically the same we had before, when there is no tooltip to show.

What do you think?

packages/stark-ui/src/modules/table/entities/column-properties.intf.ts
```diff
@@ -97,4 +97,34 @@ export interface StarkTableColumnProperties {
...
+	 * This could also be a static string with tooltip
+	 */
+	cellTooltip?: ((value: any, row?: object, columnName?: string) => string) | string;
+
+	/**
+	 * Position where the tooltip should be displayed.
+	 * Default: "below"
+	 * Possible positions: above, below, left, right, before, after
+	 */
+	cellTooltipPosition?: string;
```
christophercr, line 117: Use specific types instead of just `string` so it is clear what are the valid positions we expect :wink:

packages/stark-ui/src/modules/table/components/column.component.ts
```diff
@@ -214,6 +214,40 @@ export class StarkTableColumnComponent extends AbstractStarkUiComponent implemen
...
+	 * Position where the tooltip should be displayed.
+	 * Default: "below"
+	 * Possible positions: above, below, left, right, before, after
+	 */
+	@Input()
+	public get cellTooltipPosition(): string {
+		return this._cellTooltipPosition;
+	}
+
+	public set cellTooltipPosition(tooltip: string) {
```
christophercr, line 234: Same remark about using specific types instead of just `string`. You could re-use the ones from the ColumnProperties :wink:

packages/stark-ui/src/modules/table/components/table.component.spec.ts
```diff
@@ -128,10 +128,18 @@ describe("TableComponent", () => {
...
+	const triggerHover: Function = (element: HTMLElement): void => {
+		// more verbose way to create and trigger an event (the only way it works in IE)
+		// https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events
+		const hoverEvent: MouseEvent = document.createEvent("MouseEvent");
+		hoverEvent.initEvent("mouseenter", true, true);
+		element.dispatchEvent(hoverEvent);
+	};
+
 	const DUMMY_DATA: object[] = [
-		{ id: 1, description: "dummy 1" },
```
christophercr, line 132: Were this whitespaces actually removed by Prettier? That would be weird :thinking:

christophercr (changes requested):
Probably we need to adapt the logic to show the tooltip to avoid doing unnecessary stuff in case there is no tooltip to show...

Issue #2543, opened 2021-03-17
ui: replace pretty-data library


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

After upgrading Stark to Angular 8, we get the following error when running the showcase:

```bash
Uncaught ReferenceError: assignment to undeclared variable maxdeep
    pp pretty-data.js:54
    js pretty-data.js:334
    Webpack 14
pretty-data.js:54:2
    pp pretty-data.js:54
    js pretty-data.js:334
    Webpack 16
```

An issue about this has been opened on https://github.com/vkiryukhin/pretty-data/issues/26 and didn't receive any answer for a year. Since this library does not seem to be maintained anymore, we should investigate to replace it by another library.

## Minimal reproduction of the problem with instructions

1. Checkout master branch
2. Edit "packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts" file as following:
  ```typescript
  // Line 53 uncomment following line:
  const prettyData: any = require("pretty-data").pd;

  // +- Line 160 adapt as follows:
  this.prettyString = prettyData.xml(this.data);
  // this.prettyString = this.data;

  // +- Line 188 adapt as follows:
  this.prettyString = prettyData.sql(this.data);
  // this.prettyString = this.data;
  ```
3. Run `npm run build` && npm run update:showcase && npm run showcase

## What is the motivation / use case for changing the behavior?

Fix support for all extensions in pretty-print component.

To do so, we could use different formatters to replace ones supported by pretty-data in stark-ui:
- SQL: [sql-formatter](https://github.com/zeroturnaround/sql-formatter)
- XML: [@prettier/plugin-xml](https://github.com/prettier/plugin-xml)
- HTML: already present [prettier/parser-angular](https://prettier.io/docs/en/options.html#parser)

## Environment

<pre><code>
Angular version: 8x

</code></pre>

PR #2597, opened 2021-03-31
feat(stark-ui): replace `pretty-data` by `prettier`+`@sqltools/formatter` in pretty-print component

SuperITMan:
ISSUES CLOSED: #2543

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #2543 

Due to Angular 8 upgrade, we had to disable plugin related to `pretty-data`.

## What is the new behavior?

Remove `pretty-data` dependency and replace it by `@prettier/plugin-xml` and `@sqltools/formatter`.

```
[ ] Yes
[X] No
```

## Other information


nicanac (commented):
Nice, good job, 
juste a small question I have. For the sake of the changelog, isn't it preferable to have a second commit for the sqltools modifications ?

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
```diff
@@ -46,7 +78,8 @@ describe("PrettyPrintComponent", () => {
 		'  <menuitem value="New" onclick="CreateNewDoc()" />',
 		'  <menuitem value="Open" onclick="OpenDoc()" />',
 		'  <menuitem value="Close" onclick="CloseDoc()" />',
-		"</menu>"
+		"</menu>",
+		""
```
christophercr, line 82: Is this empty string really needed in the raw data now?

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -154,9 +158,15 @@ export class StarkPrettyPrintComponent extends AbstractStarkUiComponent implemen
 					case "html":
 						prismGrammar = Prism.languages.markup;
 						prismClass = prismClassPrefix + "markup";
-						// FIXME Adapt following line after replacing "pretty-data" dependency. See: https://github.com/NationalBankBelgium/stark/issues/2543
-						// this.prettyString = prettyData.xml(this.data);
-						this.prettyString = this.data;
+						if (this.format === "xml") {
```
christophercr, line 161: I suggest to move this block to the ´case "xml"´ above 😉

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.spec.ts
nicanac: isn't it because prettier want the file ends with newline ?

SuperITMan: @nicanac is right. Prettier end with a newline. Without this, our tests don't pass.

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
SuperITMan: I was hesitating because the logic for highlighting is the same. But you're right, it will be better for the understanding to split this since the formatter logic is not the same anymore 😊

SuperITMan:
@nicanac I think it makes to keep both in a single commit since both are replacing `pretty-data` dependency.
I added 2 dependencies to replace one:
- `@prettier/plugin-xml`
- `@sqltools/formatter`

Does it seem ok for you ? 😊

SuperITMan:
@nicanac @christophercr I adapted component code as suggested by Christopher. 
Could you have a look ? 😊

nicanac (approved):
LGTM 👍

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
```diff
@@ -151,12 +155,19 @@ export class StarkPrettyPrintComponent extends AbstractStarkUiComponent implemen
...
+						});
+						break;
+
 					case "html":
 						prismGrammar = Prism.languages.markup;
 						prismClass = prismClassPrefix + "markup";
-						// FIXME Adapt following line after replacing "pretty-data" dependency. See: https://github.com/NationalBankBelgium/stark/issues/2543
-						// this.prettyString = prettyData.xml(this.data);
-						this.prettyString = this.data;
+						this.prettyString = prettier.format(this.data, { parser: "angular", plugins: prettierPlugins });
```
christophercr, line 170: Just one thing, maybe it's worth mentioning in the changelog, with this change now the PrettyPrint component is able to format not only HTML but also Angular HTML code.

So what to do? 🤔

christophercr (commented):
LGTM... just one remark/question

packages/stark-ui/src/modules/pretty-print/components/pretty-print.component.ts
SuperITMan: Indeed, I extended the support to Angular.
I think it could be a good idea since it's an angular framework 😊
Or maybe should it be a separated feature?

It would mean changing back to html and add a new support for "angular"?

christophercr: Ah ok, well, for me it's ok to keep it, but maybe just mention this new feature in the changelog?

christophercr (approved):
Nice! :+1:

Issue #2653, opened 2021-04-28
testing: replace karma-typescript by karma +  `@angular-devkit/build-angular:karma`


SuperITMan:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead see https://github.com/NationalBankBelgium/stark/blob/master/CONTRIBUTING.md#got-a-question-or-problem
</code></pre>

## Current behavior

Stark-Testing is using karma-typescript to execute tests. 
Recently, we got troubles due to karma-typescript-es6-transform (#2513 & #2511) and we had to fix the version of the dependency.

## Expected behavior

As Angular evolved a lot since we started Stark 10, I think it's the right time to move from karma-typescript to `@angular-devkit/build-angular:karma` using `@angular-builders/custom-webpack:karma`.

## What is the motivation / use case for changing the behavior?

Improve testing maintenance and support, based on Angular configuration.

SuperITMan:
Closed by #2659

PR #2659, opened 2021-04-29
Feature/replace karma-typescript by karma

SuperITMan:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[X] CI related changes
[ ] Other... Please describe:
```

Stark-testing is using karma-typescript. 

Issue Number: #2653 

## What is the new behavior?

Using karma + `@angular-devkit/build-angular:karma` instead of karma-typescript

```
[X] Yes
[ ] No
```

## Other information


SuperITMan:
Hey @christophercr @nicanac I adapted the test logic in Stark to use karma config from angular and add our custom configuration.

In the meanwhile, I copied `stark-testing` dependencies into root package.json to be able to run `ng test stark-core` from root. The direct downside is that we have to maintain `stark-testing` dependencies manually. 
The goal will be to use `@nrwl/nx` or a similar solution in the future and to centralize dependencies in the root folder.

Could you have a look ? 😊

starter/package.json
```diff
@@ -127,8 +127,8 @@
     "@angular/platform-server": "^8.2.0",
     "@angular/router": "^8.2.0",
     "@nationalbankbelgium/code-style": "^1.5.0",
-    "@nationalbankbelgium/stark-core": "latest",
-    "@nationalbankbelgium/stark-ui": "latest",
+    "@nationalbankbelgium/stark-core": "file:../dist/packages-dist/stark-core/nationalbankbelgium-stark-core-10.2.0-71462b2f.tgz",
```
christophercr, line 130: Shouldn't these changes be reverted? Normally the Starter and Showcase should install the `latest` from all Stark packages right?

christophercr, line 130: Shouldn't these changes be reverted? Normally the Starter should install the `latest` from all Stark packages right?

starter/package.json
```diff
@@ -143,8 +143,8 @@
   },
   "devDependencies": {
     "@compodoc/compodoc": "~1.1.11",
-    "@nationalbankbelgium/stark-build": "latest",
-    "@nationalbankbelgium/stark-testing": "latest",
+    "@nationalbankbelgium/stark-build": "file:../dist/packages-dist/stark-build/nationalbankbelgium-stark-build-10.2.0-71462b2f.tgz",
```
christophercr, line 146: Shouldn't these changes be reverted? Normally the Starter should install the `latest` from all Stark packages right?

showcase/tsconfig.spec.json
```diff
@@ -1,7 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "compilerOptions": {
-    "module": "commonjs"
```
christophercr, line 4: Don't remember exactly why we supported `commonjs` libs in tests :thinking:

combine-packages-coverage.js
```diff
@@ -44,15 +44,6 @@ const nextStream = function () {
 	return fs.createReadStream(file);
 };
 
-// first prepend the base path of each package ('packages/stark-xxxx') to every source file reference in the coverage file
-// this ensures that Coveralls can show the list of packages correctly including their files
-for (const fileName of fileNames) {
-	const packageName = fileName.match(/stark-\w*/)[0];
-	const replacements = [{ searchValue: /SF:(.*)(\r|\n)/, replaceValue: `SF:packages/${packageName}/$1$2` }];
```
christophercr, line 51: Are you sure that with `ng test` the reports are split by package name? So that the coverage of all stark packages is shown in Coveralls correctly

packages/stark-rbac/base.spec.ts
```diff
@@ -18,22 +18,31 @@ if ((<any>window).NodeList && !NodeList.prototype.forEach) {
 }
 
 /* tslint:disable:no-import-side-effect */
-// FIXME: change when https://github.com/monounity/karma-typescript/issues/320 is resolved
-// tslint:disable-next-line:import-blacklist
-import "lodash-es"; // see https://github.com/monounity/karma-typescript/issues/150#issuecomment-318620280
 import "zone.js/dist/zone";
+import "zone.js/dist/zone-testing";
```
christophercr, line 22: Ah nice! Is it really that simple now?

packages/stark-testing/karma.conf.js
```diff
@@ -81,16 +54,50 @@ const rawKarmaConfig = {
...
+	restartOnFileChange: !ciDetect(),
 
 	// Timeout settings
 	browserNoActivityTimeout: 30000,
 	browserDisconnectTolerance: 1,
-	browserDisconnectTimeout: 30000
+	browserDisconnectTimeout: 30000,
+
+	// Configuration for coverage-istanbul reporter
+	coverageIstanbulReporter: {
```
christophercr, line 72: Mmm are you sure that running code coverage also in non-ci environment is fast enough? As far as I remember, this was a bit slower than normal testing without coverage, that's why we did it only for CI...

Wait... I see that the script passes `ng test --code-coverage` only for CI... so that is basically enabling code coverage only in CI right?

package.json
```diff
@@ -31,24 +31,43 @@
...
     "commitizen": "^4.2.3",
     "conventional-changelog-cli": "^2.1.1",
     "core-js": "^3.9.1",
+    "coveralls": "^3.1.0",
     "cross-env": "^7.0.3",
     "cz-customizable": "^6.3.0",
     "hammerjs": "^2.0.8",
     "husky": "^4.3.8",
+    "jasmine-core": "^3.6.0",
+    "karma": "^5.2.3",
```
christophercr, line 56: About moving all deps from `stark-testing` to the root... could we try something else so we can avoid duplication?

What about just adding `stark-testing` as DevDependency here and link it to the stark-testing folder (no need for a compiled `tgz` I think because we don't compile anything in that package). That way the deps will be installed at the root... right? :thinking:

christophercr (commented):
Nice job! Indeed, having a centralized testing logic compatible with NG CLI will always give us better compatibility with the NG ecosystem :smile: 

I just have some small thoughts about it... see my comments below.

starter/package.json
SuperITMan: Indeed, you're right. This file should not be changed. Good catch! 👍

packages/stark-testing/karma.conf.js
SuperITMan: Exactly! 😊 

The configuration for coverageIstanbulReporter is there but to enable this reporter you need the `--code-coverage` flag. It has no effect on `ng test` without the flag.

combine-packages-coverage.js
SuperITMan: Actually, since we use `ng test stark-core --code-coverage` from stark root folder instead of `ng test` from "packages/stark-core" folder, the paths are correct.

`ng test --code-coverage` from "packages/stark-core", lcov contains path "SF:src/common/error/error.ts"
`ng test stark-core --code-coverage` from root, lcov contains path "SF:packages/stark-core/src/common/error/error.ts"

package.json
SuperITMan: Ha ha, I didn't think to this since the dependencies are not well maintained after (doing `npm i` from root after upgrading deps in stark-testing has no impact on installed deps in root).

However, since we are looking for a better solution in the future to manage our monorepo, I think it's better to use your solution and not copy deps from testing into root folder 👍

SuperITMan:
@christophercr I updated my PR based on your remarks. Could you have a look ? 😊

christophercr:
@SuperITMan btw... where did the Coveralls checks go?

christophercr (approved):
Looks great!

Just the remark about Coveralls checks missing :thinking:

SuperITMan:
@christophercr It is now part of GitHub Actions:
https://github.com/NationalBankBelgium/stark/blob/fb6e448b67b21e5c5a9c52373251650900e66965/.github/workflows/build.yml#L165-L174

christophercr:
Ok, but what about the comment posted by Coveralls in the PRs to know whats the new coverage?

christophercr:
Moreover, you can see in the Coveralls page (https://coveralls.io/github/NationalBankBelgium/stark) that the latest merge you did some minutes ago about the NGRX actions is not shown in that page :thinking:

SuperITMan:
It's there again
image

https://coveralls.io/builds/39371204 :blush:

christophercr:
Alright!! Much better! 😋

nicanac:
Seems really great, I just have a concern regarding the security issue -> snyk.
I'm not well informed about this security test, because when I'm not able to get the details and the reason why it fails.

Otherwise, testing are working well on my win10 pc and on my mac Os.

nicanac (commented):
Seems really great, I just have a concern regarding the security issue -> snyk.
I'm not well informed about this security test, because when I'm not able to get the details and the reason why it fails.

Otherwise, testing are working well on my win10 pc and on my mac Os.

SuperITMan:
@nicanac 
Actually, Snyk is complaining because karma v5 and @angular-devkit/build-angular v0.803.29 contain some vulnerabilities.
Those vulnerabilities should be solved once we migrate on Angular 9 and Karma 6 😊

